PR #65, opened 2016-11-22
[sale_stock_availability] Only count stock on selected warehouse

TheCloneMaster:
with_context should be applied at product_id to be effective.  Currently it is counting from all stock locations.

Also, context should be updated to contain warehouse, if not call to super product_id_change_with_wh will also take into account all stock locations and not just selected warehouse from sale order

### DIFF
--- a/sale_stock_availability/sale_order.py
+++ b/sale_stock_availability/sale_order.py
@@ -12,9 +12,9 @@ class sale_order_line(models.Model):
     def _fnct_line_stock(self):
         available = False
         if self.order_id.state == 'draft':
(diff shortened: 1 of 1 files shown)


jjscarafia:
@nicomacr can you check this please?

thanks @TheCloneMaster

TheCloneMaster:
Hi,

Just in case, I already submited similar fix to sale_exceptions on OCA, and
they already merged...

Regards,
-Mario

nicomacr (approved):

jjscarafia:
Thanks

PR #126, opened 2018-07-09
[MOD][sale_order_type_automation] No auto-create draft invoice

rubencr7:
Auto-create invoice only when "validate automatically invoice" is checked in sale order type.


jjscarafia:
Hi @rubencr7 . Thanks for the PR. On v8 is not very clear, but if you don't want invoice to be created just don't assign journal_id.
We are not longer using v8 but I would prefer not to aprove this so that it doesn't break others.

PR #127, opened 2018-07-10
[REL] sale_require_contract to V11.

ValentinoDefelice:


sale_require_contract/data/data.xml
```diff
@@ -1,16 +1,15 @@
...
+<odoo noupdate="1">
 
-    <record id="require_contract" model="sale.exception">
+    <record id="require_contract" model="exception.rule">
       <field name="name">Contract Required</field>
       <field name="description">You must set a contract.</field>
       <field name="sequence">10</field>
       <field name="model">sale.order</field>
+      <field name="rule_group">sale</field>
       <field name="code">if not object.project_id:
```
nicomacr: Hi @ValentinoDefelice thanks for the PR
The field `project_id` not longer exist by this name, now it's call `analytic_account_id` .
Please change the field name here.
Thanks

nicomacr (changes requested):
@ValentinoDefelice 
I put a comment below

nicomacr (approved):

PR #128, opened 2018-07-10
[REL] Migration sale_quotation_products to 11.0

nicomacr:


sale_quotation_products/models/product_product.py
```diff
@@ -77,41 +83,38 @@ def write(self, vals):
...
-            lines = self.env['sale.order.line'].search([
-                ('order_id', '=', sale_order_id),
-                ('product_id', '=', self.id)])
-            self.qty = sum([self.env['product.uom']._compute_qty_obj(
-                line.product_uom,
-                line.product_uom_qty,
-                self.uom_id) for line in lines])
+        for rec in self:
+            qty = 0
+            if sale_order_id:
```
zaoral: Improves of performance of this method

1.  > if sale_order_id:

    This value is the same for all the records of the loop, so is not needed to be questioned  every time,
remove from the loop and  ask for it before the loop

2. > self.env['sale.order.line'].

   define outside the loop

Example 

```python
@api.multi
def _compute_qty(self):
    sale_order_id = self._context.get('active_id', False)
    if not sale_order_id:
        self.update({'qty': 0.0})
        return

    sale_line_obj = self.env['sale.order.line']
    for rec in self:
        lines = sale_line_obj.search([
            ('order_id', '=', sale_order_id),
            ('product_id', '=', rec.id)])
        qty = sum([line.product_uom._compute_quantity(
            line.product_uom_qty,
            rec.uom_id) for line in lines])
        rec.update({'qty': qty})
```

sale_quotation_products/models/product_product.py
```diff
@@ -77,41 +83,38 @@ def write(self, vals):
...
-                })
-            else:
-                self.env['sale.order'].browse(
-                    sale_order_id).add_products(self.id, qty)
-
-    qty = fields.Float(
-        'Quantity',
-        compute='_get_qty',
-        inverse='_set_qty')
+        for rec in self:
```
zaoral: same in this method.
1. self.env[model] outside loop
2. sale_order_id can be evaluate before the loop one time and then apply what is required.

sale_quotation_products/models/product_product.py
```diff
@@ -77,41 +83,38 @@ def write(self, vals):
             self = self.sudo()
```
zaoral: Multiple improves to this method

1.  > self = self.sudo()

    I have understood that this should not be done. Do not find the source were I read it... As far as I know should be replaced with `super(ModelClass, self.sudo()).write(vals)`

2. > vals.get('qty) and self._context.get('force_product_edit'):

    The dict.get(key) should be always used as dict.get(key, False) to return False in the key is not found and do not raise any error, also because this ones are used in a if condition.

Example

```python
  @api.multi
     def write(self, vals):
         """
         Si en vals solo viene qty y force_product_edit entonces es un dummy
         write y lo hacemos con sudo
         """
         if len(vals) == 1 and vals.get('qty', False) and self._context.get(
                 'force_product_edit', False):
             return super(ProductProduct, self.sudo()).write(vals)
         return super(ProductProduct, self).write(vals)
```

sale_quotation_products/models/product_product.py
```diff
@@ -10,6 +10,12 @@
...
 
+    qty = fields.Float(
+        'Quantity',
+        compute='_compute_qty',
+        inverse='_inverse_qty',
+    )
+
     @api.model
     def fields_view_get(self, view_id=None, view_type='form',
                         toolbar=False, submenu=False):
```
zaoral: This is a personal position: I am not really a fan about editing views from the fields_view_get method.
In this case I propose:

1. create a new editable view inherit from the original one but with primary mode one, there made the changes we need (add buttons, remove fields, set fields readonly, editable bottom, etc)
2. remove entirely the fields_view_get method
3. at view edit `Search Products` button and add to context `form_view_ref: 'module.my_new_editable_view'`

sale_quotation_products/models/product_product.py
```diff
@@ -77,41 +83,38 @@ def write(self, vals):
...
+                    lines[0].write({
+                        'product_uom_qty': qty,
+                        'product_uom': rec.uom_id.id,
+                    })
+                else:
+                    self.env['sale.order'].browse(
+                        sale_order_id).add_products(rec.id, qty)
 
     @api.multi
     def action_product_form(self):
```
zaoral: >          view_id = self.env['ir.model.data'].xmlid_to_res_id(
>            'product.product_normal_form_view')

This is the same as

> view_id = self.env.ref('product.product_normal_form_view')

Second version is simpler and easier to read

zaoral: At action_product_add_one use update instead of direct assign

sale_quotation_products/models/sale_order.py
```diff
@@ -34,8 +34,8 @@ def add_products_to_quotation(self):
     @api.multi
```
zaoral, line 39: 1. > from ast import literal_eval

use odoo save eval instead `from odoo.tools.safe_eval import safe_eval`

2. > instead of update context one by one, use only one update()
```python
   context.update(dict(
        sale_quotation_products=True,
        pricelist=self.pricelist_id.display_name,
        # we send company in context so it filters taxes
        company_id=self.company_id.id,
        partner_id=self.partner_id.id,
        search_default_location_id=self.warehouse_id.lot_stock_id.id,
         # search_default_warehouse_id=self.warehouse_id.id,
    )
```
3. > same for action_read

```python
    action_read.update(
        context=context,
        # view_mode='tree,form'.
        name=_('Quotation Products'),
    )
```

sale_quotation_products/models/sale_order.py
```diff
@@ -34,8 +34,8 @@ def add_products_to_quotation(self):
...
     def add_products(self, product_ids, qty):
         self.ensure_one()
+        sol = self.env['sale.order.line']
         for product in self.env['product.product'].browse(product_ids):
-            sol = self.env['sale.order.line']
             last_sol = sol.search(
                 [('order_id', '=', self.id)], order='sequence desc', limit=1)
             sequence = last_sol and last_sol.sequence + 1 or 10
```
zaoral, line 46:
```python
             sol = sol.new(vals)
             # we call onchange product to get required fields
             sol.product_id_change()
             # we set qty (if we set it on create odoo overwrite it to 1)
             sol.product_uom_qty = qty
             # # we call onchange qty)
             sol.product_uom_change()
             # we convert to write
             vals = sol._convert_to_write(sol._cache)
             sol.create(vals)
```

remove no needed intra lines coments

zaoral (changes requested):
Hi @nicomacr 
Technical review ready

zaoral (approved):

PR #130, opened 2018-07-11
[REL] Migration sale_contract_restrict_domain to 11.0

nicomacr:


sale_ux/models/sale_order.py
```diff
@@ -22,6 +22,13 @@ class SaleOrder(models.Model):
         track_visibility='onchange',
         copy=False,
     )
+    commercial_partner_id = fields.Many2one(
+        'res.partner',
+        string='Commercial Entity',
+        related='partner_id.commercial_partner_id',
+        store=True,
```
zaoral, line 29: Dont need to store the related field to make it searchable,
Only need to to add a search method instead: this way we are saving a lot of database space

```diff
- store=True,
+ search='_search_commercial_partner_id',
```

The search method is like:

```python
     @api.model
     def _search_commercial_partner_id(self, operator, operand):
         return [('partner_id.commercial_partner_id', operator, operand)]
```

sale_ux/views/sale_order_views.xml
```diff
@@ -21,6 +21,12 @@
             <button name="action_cancel" position="attributes">
                 <attribute name="states">draft,sent,sale,done</attribute>
             </button>
+            <field name="partner_id" position="after">
+                <field name="commercial_partner_id" invisible="1"/>
+            </field>
+            <field name="analytic_account_id" position="attributes">
+                <attribute name="domain">['|',('partner_id','=',False),('partner_id.commercial_partner_id','=',commercial_partner_id)]</attribute>
```
zaoral: Considere when the analytic account has not partner and when it have the commercial_partner_id but currently not taking into account the partner itself. 

Maybe you could use the next domain

 > ['|','|',('partner_id','=',False), ('partner_id','=', partner_id), 
 ('partner_id.commercial_partner_id','=',commercial_partner_id)]

sale_ux/README.rst
```diff
@@ -25,6 +25,7 @@ Several Improvements to sales:
 #. Add button to force "invoiced" only for admin with tec features.
 #. Add tracking to payment term and pricelist on sale orders.
 #. Add option to show SO reference field on tree view and on main section of form view.
+#. Add filters contracts on sales orders so you can only select the ones with same partner as the sales order.
```
zaoral: Update doc, not mention contract, mention both partner/commercial partner

```diff
-#. Add filters contracts on sales orders so you can only select the ones with same partner as the sales order.
+#. Add filter to be able to select analytic account on sale orders with the same partner/commercial partner as the sale order.  
```
Add new item

#. Add Commercial Partner is automatic set in the sale order taking into account the partner configuration

zaoral (changes requested):
Hi @nicomacr 

Technical review ready
Best Regards

sale_ux/views/sale_order_views.xml
jjscarafia: If partner is commercial_partner_id then no problem. partner_id.commercial_partner_id will be the same as partner_id. 
So, default domain is ok "sale_usability_delivery a sale_delivery_ux"

sale_ux/models/sale_order.py
jjscarafia: To be searchable, if I'm not wrong, related fields don't need "search", they are searchable by default. From v9+ I think that you can also group by a related field without making it stored, So I beleieve that just removing  store=True is enaught (no need for search method)

PR #141, opened 2018-07-25
[REL] sale_order_type_user_default to V11.

ValentinoDefelice:
[FIX] Update README.


sale_order_type_user_default/models/sale_order.py
```diff
@@ -15,4 +15,5 @@ def _get_order_type(self):
 
     type_id = fields.Many2one(
         'sale.order.type',
-        default=_get_order_type)
+        default=_get_order_type,
```
nicomacr: @ValentinoDefelice this default should be this way :
`  default=lambda so: so._get_order_type()`

sale_order_type_user_default/models/res_users.py
```diff
@@ -7,5 +7,5 @@ class ResUsers(models.Model):
```
nicomacr, line 7: to this file run `chmod -x res_users.py`

sale_order_type_user_default/models/sale_order.py
```diff
@@ -15,4 +15,5 @@ def _get_order_type(self):
```
nicomacr, line 15: to this file run `chmod -x sale_order.py`

PR #233, opened 2019-07-16
[REL] sale_ux migration to 12.0

nicomacr:
[REM] remove these views: product_product_views.xml, product_template_views.xml now are not necesary.
[FIX] change to instalable the module
[FIX] some change in view adapting to new position elements
[FIX] use the block confim to sale order using by odoo and add new state 'sale'


zaoral (approved):
+1!

PR #243, opened 2019-07-26
[REL] sale exceptions print migration to 12.0

ValentinoDefelice:


sale_exception_print/models/sale_order.py
```diff
@@ -40,7 +40,7 @@ def detect_print_exceptions(self):
         for order in self:
```
nicomacr: do `with_context(print_exceptions=True)` here not inside of the iteration.

nicomacr (approved):

PR #253, opened 2019-08-06
[REL] Migration sale_stock_ux to 12.0

nicomacr:


sale_stock_ux/models/sale_order_line.py
```diff
@@ -23,8 +23,8 @@ class SaleOrderLine(models.Model):
     # TODO This should be computed field
```
zaoral: remove comment, you already change the field to computed field

sale_stock_ux/models/sale_order_line.py
```diff
@@ -23,8 +23,8 @@ class SaleOrderLine(models.Model):
...
     qty_returned = fields.Float(
         string='Returned',
+        compute='_compute_qty_returned',
         copy=False,
-        default=0.0,
         digits=dp.get_precision('Product Unit of Measure'),
         readonly=True,
```
zaoral: I think that this readonly is not necessary since now is a compute fieñd

zaoral (commented):
LGTM +1 @nicomacr 

I leave you here some comments, you decide if you will like to apply it:

* I think by that maybe the field all_qty_delivered could be removed
* remove string='Delivery Status',  of delivery_field in the sale.order.line already same as name of the field
* rename file stock_return_picking_line_views.xml to  stock_return_picking_views.xml

Best Regards

zaoral:
note: I was not able to make functional test, runbot is failling because the product_ux module is missing (maybe has not been migrated yet)

Please if you could make a rebase will be great

nicomacr:
@zaoral the changes suggested has been apply

zaoral (approved):
Excellent! @nicomacr 

There is a yellow runbot, it seems like is because of multiple warnings, ones related to this and other modules.

Please could you review if it is possible to fix them?
Best Regards

PR #311, opened 2020-03-11
[MIG] crm_ux: Migration to 13.0

daniellibonati:
* Migrated with odoo-module-migrate
* Adjusted xpath to match right website field
* Adjusted inheritance


sotolucas (changes requested):
Seems to me that this module is not longer needed.
Check [this](https://github.com/odoo/odoo/blob/adb2186a8cb2c411d5fc04bc4dfb1998424909fa/addons/crm/views/crm_lead_views.xml#L230). 
Also I've done tests and company_id field is shown by default.

sotolucas (approved):
LGTM now :+1:

PR #395, opened 2021-01-07
[ADD] sale_ux: products images on reports

PabloPaezSheridan:


PabloPaezSheridan:
Incompleto, no funcional

PR #416, opened 2021-02-05
[13.0] [PORT] from 12.0

nicomacr:
[12.0] [FIX] portal_sale_distributor: Avoid problems with the comercial user when create SO.

After this odoo change https://github.com/odoo/odoo/commit/99d0308d5870088970667744fdd0c818299927bc
This hack is necessary in the method "message auto subscribe" to make it's works.


jcadhoc (dismissed):

jcadhoc (approved):
Tested and works fine

PR #455, opened 2021-10-13
[13.0] [FIX] sale_three_discounts: Only compute on customer invoices

nicomacr:


zaoral (dismissed):
+1 @nicomacr 

Note: We can also the method the [is_sale_document()](https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L1879-L1884 
) method , let me know what do you think?

nicomacr:
@zaoral excellent, i think is better this way!! 
Changes applied.

zaoral (approved):

PR #548, opened 2022-06-14
[FIX] sale_ux: create advance invoice from SO

jjscarafia:


roboadhoc:
[Pull request status dashboard](https://runbot.adhoc.com.ar/ingadhoc/sale/pull/548).

PR #1186, opened 2025-04-07
[FIX] sale_ux: Fix in method calls. We need to call _compute_tax_id instead of _compute_tax_ids

rov-adhoc:
The method `_compute_tax_ids` was incorrectly used in the context where the correct method is `_compute_tax_id`. 

This fix ensures that the correct tax computation method is invoked, 
leading to accurate tax calculations.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1186).

cav-adhoc:
@roboadhoc r+

PR #1213, opened 2025-04-21
[FIX] sale_order_type_ux: Fix downpayment creation with multicompany sale order type

rov-adhoc:
When creating a downpayment invoice from a sale order using a sale order type with a journal from a different company, the taxes were not correctly adapted to the journal's company. This caused tax inconsistencies and incorrect tax mapping.

This fix ensures that:
- The invoice is prepared in the correct company context based on the journal set on the sale order type.
- Downpayment lines use taxes correctly mapped to the company issuing the invoice.
- If a fiscal position is defined, the taxes are mapped accordingly.

This avoids tax company mismatch errors in multicompany environments.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1213).

matiasperalta1:
@roboadhoc r+

PR #1224, opened 2025-04-30
[FIX] sale_ux: Refine sale order cancellation check based on invoice and debit note payment status

rov-adhoc:
Improves the condition that prevents a sale order from being cancelled if related invoices or debit notes are still active.

- For customer invoices (`out_invoice`), skip those that are already reversed.
- For customer debit notes (`out_debit`), skip those that are fully paid.
- Keeps existing exclusions for moves in draft or cancelled state.

This prevents false positives in the cancellation restriction and ensures only truly active financial documents block the operation.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1224).

lav-adhoc:
@roboadhoc r+

PR #1287, opened 2025-07-18
[FIX] sale_order_type_ux: browse fiscal position from second company

cav-adhoc:
The original code checked if the fiscal position's company was different from the sales order's company. If the second company (the invoice company) did not have a fiscal position, nothing was assigned.

I added a condition so that if no fiscal position is defined, it will always look for the fiscal position of the invoice company.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1287).

rov-adhoc:
@roboadhoc r+

PR #1293, opened 2025-07-22
[FIX] sale_three_discounts: Filter out invoice and refund types in discount computation

rov-adhoc:
When discount is set from a purchase order, the system should not replace with the discount computation.

This code was in 16 and deleted in migration of 17 by mistake.

https://github.com/ingadhoc/sale/blob/966f3f2d7b1cae18df45dc2508d70e2b805912db/sale_three_discounts/models/account_invoice_line.py#L72C9-L72C92


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1293).

cav-adhoc:
@roboadhoc r+

PR #1307, opened 2025-07-30
[FIX] portal_sale_distributor: fix en s.o name

lav-adhoc:
The problem is that when a portal distributor user creates a sales order in the second company, the order name appears as 'New' in the portal view, even though the correct name is assigned in the backend.

With this method, we ensure that the correct name is displayed.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1307).

PR #1327, opened 2025-08-29
[FIX] sale_order_type_ux: restrict sale order type to active company journals

rov-adhoc:
When computing `sale_type_id`, the search could return a `sale.order.type` whose `journal_id` belongs to a different company that could not be selected. This caused wrong assignments when trying to create a new customer invoices.

The fix ensures that only sale order types with journals from the current record's company (or no journal at all) are considered.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1327)

maq-adhoc:
@roboadhoc r+

PR #1531, opened 2026-01-12
[FIX] sale_order_type_automation: correctly set sale_type_id when running automation of invoicing instead of using default sale type id that is not correct when the sale already has one defined

rov-adhoc:


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1531)

cav-adhoc:
@roboadhoc r+

PR #1637, opened 2026-04-20
[ADD] portal_sale_distributor_*: fix distributor portal access errors

gal-adhoc:
Fixes the errors a distributor portal user hits on a sale order, reported in task #65950, and adds the Update Validity button they were missing.

## What was broken

Every failure had its own cause, they only looked like one bug because they all surfaced as an access error while saving an order:

- **Choosing a customer and confirming an order.** A distributor is a portal user, so `base.res_partner_portal_public_rule` limits them to their own commercial family. Saving an order makes the ORM read the contact of the company, which falls outside of it.
- **Opening the product catalog.** `sale_project` restricts the task fields of `sale.order` to project users, and `project.task` had no access at all.
- **Adding a combo product.** Upstream bug, see below.
- **Editing a confirmed order.** This was never blocked by a business rule: `sale_stock` reads `stock.reference` through `sale_order_stack_reference_ids` when saving the line, and the distributor had no access to that model. With it granted they change quantities like an internal user does. There is no message to show, the operation simply works.

## Decisions worth reviewing

**The `res.partner` rule is deliberately one term.** Record rules of the groups a user belongs to are OR-ed, so the new rule only has to add what the core portal rule leaves out, which is the contact of the company. It started as `[(1, '=', 1)]` and then as a three term domain including `('partner_share', '=', False)`; both handed the distributor far more than they need — the second one exposed the contact of every internal user. Measured on a test database, the current domain takes the partners a distributor can read from 46 down to 2, with the whole flow still working. Every term added here is readable across the database, so it should only be widened against a reproduced access error. The rule lives outside of the `noupdate` security file so the domain can still be tightened later.

**`sale_margin` is left alone.** An earlier version of this branch extended the groups of `margin`, `margin_percent` and `purchase_price`, which would have exposed the cost to the distributor. #1811 already covers the same error by creating the line as superuser and keeping the fields out of reach, so the bridge module for it was dropped. The order level totals block is hidden in `_get_view` instead: `sale.order.margin` carries no groups of its own, unlike its line level counterparts, so nothing kept it out of the portal form. Note this only hides it, the value stays readable over rpc.

**`update_date_prices_and_validity` checks access before escalating.** It is a public method, so it can be called over rpc with any id; the sudo alone would have let a distributor rewrite the date and the prices of somebody else's order.

## Upstream bug with combo products

Adding a combo broke the save with `Expected singleton: sale.order.line()`. While the order is created the combo item line points at its parent through `linked_virtual_id`, but the parent has no id yet, so core's `_get_linked_line()` calls `ensure_one()` on an empty recordset. The web client normally hides this because it sends `discount` in the values and the field is then not precomputed; a distributor has no `sale.group_discount_per_so_line`, so `discount` is not in their view, never sent, and the precompute runs. Reproduced in a shell with an internal user too, by leaving `discount` out. Only the backend form reaches it — the website cart links combo lines through `linked_line_id`, on a parent that already exists.

`_get_linked_line()` now returns an empty recordset instead of raising, scoped through a context flag to the create that reaches it.

**Known gap:** a combo whose parent carries a discount saves its items at 0, because `discount` does not depend on `linked_line_id` and nothing recomputes it once the parent exists. Core leaves an internal user in the same spot, so it is being fixed apart from this PR.

## Also

- `portal_sale_distributor_project`: read access to `project.task` and to the task fields of `sale.order`
- `portal_sale_distributor_sale_order_validity`: Update Validity button next to Update Prices, hidden at its native position for distributors
- `portal_sale_distributor_loyalty`: read access to `loyalty.program` and `sale.order.coupon.points`
- `name_search` on `res.partner` narrows the customer dropdown to the commercial family. It carries `@api.model`, without it `call_kw` breaks every partner autocomplete for every user.
- READMEs for the two new modules

## Tested

Distributor portal user on a test database: choosing a customer, saving and confirming with and without lines, product catalog, combo and storable products, duplicating an order, editing a confirmed one, loyalty, and the chatter after confirming. Cross checked with an internal user that margin and cost stay visible for them and the Update Validity button keeps its native position.

## Related

Task: #65950
Follows #1811, which fixes two of these through a different route.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1637)

portal_sale_distributor_project/models/sale_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    reinvoiced_sale_order_id = fields.Many2one(
+        groups="sales_team.group_sale_salesman,portal_sale_distributor.group_portal_backend_distributor"
+    )
+
+
+class SaleOrder(models.Model):
```
lef-adhoc, line 8: Queres separar acá una clase por archivo ?

portal_sale_distributor/security/portal_sale_distributor_security.xml
```diff
@@ -58,4 +58,15 @@
         <field eval="0" name="perm_unlink"/>
     </record>
 
+    <record id="portal_res_partner_distributor_rule" model="ir.rule">
```
lef-adhoc: No entendí muy bien por qué es esto, no es mejor agregar un permiso de acceso antes que una regla? Las reglas generalmente las usamos para restingir con algún dominio en específico pero acá tiene [(1, '=', 1)]

portal_sale_distributor/models/partner.py
```diff
@@ -9,3 +10,22 @@ class ResPartner(models.Model):
     credit_limit = fields.Float(
         groups="account.group_account_invoice,account.group_account_readonly,portal_sale_distributor.group_portal_backend_distributor"
     )
+
+    def name_search(self, name="", domain=None, operator="ilike", limit=100):
+        # El grupo distribuidor tiene una ir.rule de lectura permisiva sobre
```
lef-adhoc: Eliminaría este comentario porque no suma mucho

portal_sale_distributor/models/sale_order.py
```diff
@@ -103,3 +106,16 @@ def _get_view_cache_key(self, view_id=None, view_type="form", **options):
     def action_update_prices(self):
         self = self.sudo()
         super().action_update_prices()
+
+    def _update_order_line_info(self, *args, **kwargs):
+        if self.state == "sale" and self.env.user.has_group("portal_sale_distributor.group_portal_backend_distributor"):
+            try:
+                return super()._update_order_line_info(*args, **kwargs)
+            except AccessError:
+                raise UserError(
```
lef-adhoc: Acá mejor dejar el mensaje en inglés y con la nueva forma de traducción de v19: self.env._("mensaje")

portal_sale_distributor_sale_order_validity/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+# directory
+##############################################################################
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def update_date_prices_and_validity(self):
+        self = self.sudo()
```
lef-adhoc: Pondría que sea sudo solo es es portal distributor:
if self.env.user.has_group("portal_sale_distributor.group_portal_backend_distributor"):
    self = self.sudo()

portal_sale_distributor/models/sale_order_line.py
```diff
@@ -19,6 +19,20 @@ def create(self, vals_list):
             return super(SaleOrderLine, self.sudo()).create(vals_list).with_env(self.env)
         return super().create(vals_list)
 
+    def _get_linked_line(self):
```
lef-adhoc, line 23: Acá estas cambiando todo el comportamiento del get_linked_line sin importar si es distributor o no

Haría algo como 

if self.env.user.has_group("portal_sale_distributor.group_portal_backend_distributor"):
    ... (lo que deberia hacer)
else:
    return super()._get_linked_line()

portal_sale_distributor_sale_order_validity/views/sale_order_view.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//button[@name='update_date_prices_and_validity']" position="attributes">
+                <attribute name="groups">!portal_sale_distributor.group_portal_backend_distributor</attribute>
+            </xpath>
+            <group name="order_details" position="inside">
+                <div groups="portal_sale_distributor.group_portal_backend_distributor">
+                    <button
+                        name="update_date_prices_and_validity"
+                        type="object"
+                        string=" Update Validity"
```
lef-adhoc: Hay un espacio de más al inicio

portal_sale_distributor/security/res_partner_security.xml
```diff
@@ -0,0 +1,27 @@
...
+         their own commercial family. That is the right business boundary, but saving or
+         confirming an order makes the ORM read partners outside of it (the company contact, the
+         salesman, followers) and it raises an access error. Record rules of the groups a user
+         belongs to are OR-ed, so this one adds those partners back on top of the core rule
+         instead of replacing it. Keep it as narrow as possible: a [(1, '=', 1)] domain here
+         exposes every contact in the database to the distributor. -->
+    <record id="portal_res_partner_distributor_rule" model="ir.rule">
+        <field name="name">res.partner: read access for distributor</field>
+        <field name="model_id" ref="base.model_res_partner"/>
+        <field name="domain_force">['|', '|',
```
lef-adhoc: Acá habría que revisarlo nuevmaente porque ampliaste mucho los permisos del portal distribuidor al hacerlo con OR  -> '|'

Osea al tener esta condición en OR ('partner_share', '=', False) estas diciendo que puede ver a todos los usuarios internos el portal distribuidor y eso no estaría correcto

portal_sale_distributor_sale_order_validity/models/sale_order.py
gal-adhoc: Hecho!

lef-adhoc:
@roboadhoc r+ bump

PR #1640, opened 2026-04-21
[FIX] sale_ux: keep subcontact pricelist inherited from parent on create

gal-adhoc:
> 🎫 **Helpdesk ticket** #115317

## Changes

- [FIX] sale_ux: keep subcontact pricelist inherited from parent on create

## Problem

When creating a subcontact, Odoo's core _commercial_fields mechanism correctly syncs specific_property_product_pricelist from the parent company. However, sale_ux's create override was immediately overwriting it with None, causing the subcontact to fall back to the first pricelist in sequence instead of inheriting the parent's pricelist.

## Solution

Skip the override for partners that have a parent_id. Their pricelist is already managed by Odoo's commercial-fields sync and must not be touched.

## Related

Helpdesk ticket: #115317
Branch: 18.0-h-115317-gal

### DIFF
--- a/sale_ux/models/res_partner.py
+++ b/sale_ux/models/res_partner.py
@@ -17,6 +17,11 @@ class ResPartner(models.Model):
         pricelist = self.env["product.pricelist"].search([], limit=1, order="sequence")
         for partner, vals in zip(partners, vals_list):
             if "specific_property_product_pricelist" not in vals:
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1640)

lef-adhoc:
@roboadhoc r+ nobump

PR #1641, opened 2026-04-21
[FIX] sale_order_type_automation: check type_id is set before checking invoicing automation

gal-adhoc:
## Changes

- [FIX] sale_order_type_automation: check type_id is set before checking invoicing automation

Without this check, a sale order with no type_id would pass the filter in `run_invoicing_atomation` because accessing `invoicing_atomation` on an empty Many2one returns `False`, and `False != "none"` evaluates to `True`, causing the order to be auto-invoiced unintentionally.

## Related

Helpdesk ticket: #115934
Branch: `18.0-h-115934-gal`

### DIFF
--- a/sale_order_type_automation/models/sale_order.py
+++ b/sale_order_type_automation/models/sale_order.py
@@ -17,7 +17,8 @@ class SaleOrder(models.Model):
     def run_invoicing_atomation(self):
         for rec in self.filtered(
             lambda x: (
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1641)

lef-adhoc:
@roboadhoc r+ nobump

PR #1659, opened 2026-05-04
[FIX] sale_three_discounts: preserve manual discounts when confirming order

gal-adhoc:
> 🎫 **Helpdesk ticket** #116714

## Changes

- [FIX] sale_three_discounts: preserve manual discounts when confirming order

## Root cause

When confirming a sale order, `sale_loyalty.action_confirm()` calls `_update_programs_and_rewards()` which unlinks loyalty cards. That triggers an ORM `flush_all`, recomputing `_compute_discount()` while the order is still in `draft` state. Since `lines_to_update` was non-empty at that point, the condition reset `discount1/2/3` to 0 before the state transitioned to `sale`.

## Fix

- `sale_order.py`: override `action_confirm()` to inject `confirming_order=True` in the context before calling `super()`.
- `sale_order_line.py`: guard the discount reset with `not confirming_order`, so the reset only runs outside of the confirmation flow (or when `recompute_prices=True`, i.e. the user explicitly clicks "Update Prices").

## Related

Helpdesk ticket: #116714
Branch: `18.0-h-116714-gal`

### DIFF
--- a/sale_three_discounts/models/sale_order.py
+++ b/sale_three_discounts/models/sale_order.py
@@ -8,5 +8,8 @@ from odoo import models
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    def action_confirm(self):
+        return super(SaleOrder, self.with_context(confirming_order=True)).action_confirm()
+
     def _recompute_prices(self):
         super(SaleOrder, self.with_context(recompute_prices=True))._recompute_prices()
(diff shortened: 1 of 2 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1659)

lef-adhoc:
@roboadhoc r+ nobump

PR #1667, opened 2026-05-08
[fix] sale_stock_ux: Skip returned quantity deduction for subscriptio…

maq-adhoc:
…n procurements

When invoicing subscriptions with previous returns, the system was deducting quantity_returned from product_qty, resulting in negative quantities. Odoo's automatic reversal then inverted locations and picking_type, creating IN (receipts) instead of OUT (deliveries).

Added _check_is_recurring_invoice() validation to exclude subscriptions from this deduction, as their returned quantities are already accounted for in qty_delivered by sale_subscription_stock.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1667)

lef-adhoc:
@roboadhoc r+ nobump

PR #1724, opened 2026-06-22
[FIX] sale_stock_ux: keep return/exchange move demand correct

jcadhoc:
## Problem

On a delivered sale order, after a first return **with refund**
(`to_refund=True`, so `quantity_returned > 0`), a second operation on the same
SO produced a wrong demand on the newly created moves:

- a second **Return** move was born with `quantity - quantity_returned`
- a **Return for exchange** (both the incoming return and the re-delivery move)
  was born with `quantity - quantity_returned`

So a 3-unit return/exchange showed demand `1` instead of `3`, and when
`quantity_returned >= requested` the demand became `0`/negative, blocking the
operation entirely.

## Cause

The `stock.move.create()` hack in `sale_stock_ux` subtracts `quantity_returned`
from every new move carrying a `sale_line_id`. The docstring already flagged it:
`TODO: Solo deberia impactar en movimientos de salida`. Return moves
(`origin_returned_move_id`) and exchange moves (`is_exchange_move`) were being
discounted too.

## Fix

Scope the discount to genuine outgoing deliveries by excluding return and
exchange moves:

```python
if vals.get("sale_line_id") and not vals.get("is_exchange_move") \
        and not vals.get("origin_returned_move_id"):
```

The behaviour for real new deliveries (e.g. increasing the ordered qty after a
refund return) is left untouched.

## Tests

`sale_stock_ux/tests/test_return_exchange_demand.py`:

- `test_second_return_demand` — refund return + second return
- `test_return_then_exchange_demand` — refund return + exchange

Both fail on the previous code (`1.0 != 3.0`) and pass with the fix.

## Test plan

1. SO of 10 storable units, confirm, deliver 10.
2. Return 2 with **Refund** checked; validate.
3. Return / Return for exchange of 3 from the same delivery → demand is `3`.

Internal ref: helpdesk ticket #120871.

### DIFF
--- a/sale_stock_ux/models/stock_move.py
+++ b/sale_stock_ux/models/stock_move.py
@@ -21,14 +21,24 @@ class StockMove(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        """La cantidad devuelta no deberia ser contada en nuevos movimientos de stock
-        que se creen a partir de una orden de venta.
+        """La cantidad devuelta no deberia ser contada en los nuevos movimientos de
+        entrega genuinos que se creen a partir de una orden de venta.
         Agregamos un HACK para que si esta instalado secondary unit sea recomputada en la creacion de
         movimientos de stock.
-        TODO: Solo deberia impactar en movimientos de salida (uno o mas pasos)
+
+        Solo debe impactar en entregas genuinas: NO en los movimientos de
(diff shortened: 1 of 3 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1724)

mav-adhoc:
@roboadhoc r+ bump

PR #1732, opened 2026-06-25
[IMP] sale_price_checker: cache pricelists-enabled check on res.company

lef-adhoc:
Move the Pricelists-feature check out of the controller into a res.company method decorated with @ormcache(cache="groups"), so it is computed once instead of on every barcode scan. The 'groups' cache group is invalidated automatically (across workers) whenever group implications change -- i.e. when the feature is toggled in Sales > Configuration.

### DIFF
--- a/sale_price_checker/controllers/main.py
+++ b/sale_price_checker/controllers/main.py
@@ -8,18 +8,6 @@ from odoo.tools.misc import format_amount
 
 
 class PriceCheckerController(http.Controller):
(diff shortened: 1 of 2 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1732)

gal-adhoc:
@roboadhoc r+ nobump

PR #1733, opened 2026-06-25
[IMP] sale_ux: use <setting> tag for searchable settings in Sales config

gal-adhoc:
## Summary

The Sales configuration screen (`res.config.settings`) has a built-in search bar. Odoo's search engine only indexes and filters records that use the `<setting>` component (rendered as `SearchableSetting`) — it does **not** index standalone `<div>` elements inserted directly inside a `<block>`.

In `sale_ux`, 8 settings were defined as plain `<div class="mt16">` blocks inside the `quotation_order_setting_container` block. Their labels and descriptions were completely invisible to the search engine, so typing terms like "show", "salesman", or "pricelist" would not surface them.

This PR converts all 8 to proper `<setting>` elements with `help=` attributes (which map to the muted description text), making them fully searchable.

## Changes

- `sale_ux/views/res_config_settings_views.xml`: convert 8 `<div class="mt16">` settings to `<setting>` elements with `id` and `help`.
  - `group_allow_any_user_as_salesman`
  - `group_sale_reference_on_tree_and_main_form`
  - `update_prices_automatically` (also adds `string=` since the field has none)
  - `move_internal_notes`
  - `move_note`
  - `show_product_image_on_report` (preserves the conditional `product_image_size` sub-block)
  - `analytic_account_without_company`
  - `auto_select_all_documents`

The two existing `<setting>` elements in the same view (`cancel_old_quotations`, `group_delivery_date`) were already correct and are not touched.

## Test plan

- Open **Settings → Sales**, use the search bar.
- `show` → finds "Show Product Image", "Show client reference" (Customer reference…).
- `salesman` → finds "Allow any user as salesman".
- `pricelist` or `update prices` → finds "Update Prices Automatically".
- `notas` / `términos` → finds move_internal_notes / move_note.
- When a term matches, unrelated settings in the same block are hidden (previously the plain `<div>` elements were always visible regardless of the search query).
- Enabling "Show Product Image" still reveals the `product_image_size` selector. Save/reload preserves values.

Internal reference: https://www.adhoc.inc/odoo/knowledge/12314/project.task/69857

### DIFF
--- a/sale_ux/views/res_config_settings_views.xml
+++ b/sale_ux/views/res_config_settings_views.xml
@@ -6,84 +6,45 @@
         <field name="arch" type="xml">
 
             <xpath expr="//block[@name='quotation_order_setting_container']" position="inside">
-                <div class="mt16">
-                    <div>
-                        <field name="group_allow_any_user_as_salesman"/>
-                        <label for="group_allow_any_user_as_salesman"/>
-                    </div>
-                    <div class="text-muted">
-                        By default, Odoo only allows to choose users who have sales permission in sales orders and only internal users in the sales' teams . This option enables to choose, in both places, users of any type (internal or portal)
-                    </div>
-                </div>
-                <div class="mt16">
-                    <div >
-                        <field name="group_sale_reference_on_tree_and_main_form"/>
-                        <label for="group_sale_reference_on_tree_and_main_form"/>
-                    </div>
-                    <div class="text-muted">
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1733)

lef-adhoc:
@roboadhoc r+ bump

PR #1738, opened 2026-06-30
[FIX] sale_gathering: exclude cancelled down payments from gathering …

lef-adhoc:
…balance

A cancelled down payment line was still counted in gathering_balance, inflating the balance (and withdrawn_amount). Skip lines whose down payment state is 'cancel' and depend on order_line.invoice_lines.parent_state so the balance recomputes when the related down payment invoice is cancelled.

Partial backport of ingadhoc/sale#1730 (19.0) to 18.0.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1738)

gal-adhoc:
@roboadhoc r+ nobump

PR #1742, opened 2026-07-01
[FIX] sale_gathering: count only paid down payments in gathering balance

lef-adhoc:
The previous fix excluded lines with dp_state == 'cancel', but when a duplicate anticipo line exists with mixed invoice states (e.g. a shared posted canje invoice plus a never-posted draft anticipo), _get_downpayment_state() returns '' for both lines and neither gets excluded.

Replace the dp_state check with a payment-based criterion: a down payment line counts in gathering_balance only if it has at least one out_invoice whose payment_state is 'paid' or 'in_payment'. This correctly excludes phantom lines whose original anticipo was never posted/paid, while including legitimate paid anticios regardless of any canje/NC also linked.

Update the depends accordingly to track payment_state changes.

Closes #115776

### DIFF
--- a/sale_gathering/models/sale_order.py
+++ b/sale_gathering/models/sale_order.py
@@ -45,7 +45,15 @@ class SaleOrder(models.Model):
         for order in orders_gathering:
             lines = order._get_gathering_lines()
             total_downpayment_amount = 0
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1742)

gal-adhoc:
@roboadhoc r+ nobump

PR #1747, opened 2026-07-02
[IMP] sale_ux: make order line list page size configurable

les-adhoc:
## What
Make the order line list page size configurable on the sale order form.

## Why
Same rationale as the purchase side: Odoo core hardcodes `limit="200"` on the `order_line` `<list>`, so orders with many lines load all of them at once and slow the browser render. This adds an admin setting to restore pagination; default-off (empty/0 keeps the core behavior).

## How
- New setting **Lines per page** under *Settings → Sales* backed by `ir.config_parameter` `sale_ux.order_line_view_limit`.
- `sale.order._get_view` sets `limit` on the `order_line` list when the parameter is > 0; `_get_view_cache_key` folds the value into the form view cache key so changing the setting invalidates the cached arch.
- Empty / 0 / non-numeric / negative → no change, core default (200) is kept.

## Test plan
1. Upgrade `sale_ux`.
2. *Settings → Sales → Lines per page* = 40; open a sale order with > 200 lines → lines paginate at 40.
3. Clear the setting → 200 without pagination (core default).
4. Set a non-numeric or negative value on the System Parameter → views still load, value is treated as 0.

Companion PR (same ticket): ingadhoc/purchase#346 — mirrors the same change on purchase orders.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1747)

PR #1748, opened 2026-07-02
[FIX] sale_order_type_automation: recover aborted transaction on auto…

lef-adhoc:
…mation failure

When action_post (or the automatic payment registration) fails with a database-level error that aborts the transaction -- e.g. the AR e-invoice code commits the CAE and discards the savepoint opened around action_post, and a later error leaves the cursor aborted -- posting the failure notice over the broken cursor raised InFailedSqlTransaction, tearing down the whole order confirmation and masking the real error.

Recover the transaction before posting the failure notice on the invoice and the sale order, so the confirmation is not torn down and the real error is not masked. The savepoint behaviour for recoverable errors (invoice kept in draft, error logged on invoice and sale order) is preserved.

Tickets: 121857, 121927

### DIFF
--- a/sale_order_type_automation/models/account_move.py
+++ b/sale_order_type_automation/models/account_move.py
@@ -2,7 +2,12 @@
 # For copyright and license notices, see __manifest__.py file in module root
 # directory
 ##############################################################################
+import logging
+
+import psycopg2
 from odoo import fields, models
+
+_logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1748)

gal-adhoc:
@roboadhoc r+ nobump

PR #1750, opened 2026-07-03
[IMP] sale_ux: prevent re-confirming an already confirmed sale order

lef-adhoc:
Raise a clear UserError when action_confirm is called on a sale order already in 'sale' state, instead of the generic native message. Add Spanish translation, README entry and test.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1750)

PR #1752, opened 2026-07-03
[FIX] sale_gathering_automation: pass first_gathering_invoice context…

lef-adhoc:
… when creating gathering invoice

When confirming a gathering sale order with automation, _create_invoices was called without the first_gathering_invoice=True context. This caused _prepare_base_line_for_taxes_computation to use product_uom_qty (already 0 after _action_confirm) instead of initial_qty_gathered, resulting in an empty gathering invoice that failed automatic validation.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1752)

gal-adhoc:
@roboadhoc r+ nobump

PR #1756, opened 2026-07-06
[IMP] sale_loyalty_ux: block loyalty domains using '=' against a list

lef-adhoc:
Domains stored on loyalty programs/rules from the web builder using the '=' (or '!=') operator against a list value, e.g.
[('partner_id.category_id', '=', [21])], make Odoo's domain parser log a warning on every sale order recompute, flooding production logs.

Add an @api.constrains on loyalty.program.sale_domain and loyalty.rule.product_domain that walks the domain with odoo.fields.Domain.iter_conditions() and blocks saving such a domain, forcing the 'in' / 'not in' operator instead.

Task #70341


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1756)

PR #1762, opened 2026-07-10
[IMP] sale_stock_ux: batch stock-by-location and reserved qty computes

les-adhoc:
## What
Batch two per-line computed fields on `sale.order.line` that were causing an N+1 on the sale order form load.

## Why
Opening a sale order with many lines was slow. Profiling on a real order showed that the stock-info popover fed two non-stored computes that each hit `stock.quant` **once per line**:
- `stock_by_location` — a `read_group` per line.
- `total_reserved_quantity` — a `search` per line.

## How
- `_compute_stock_by_location`: a single `_read_group` grouped by `product_id, location_id` for the whole recordset, distributed in Python.
- `_compute_total_reserved_quantity`: one `_read_group` per distinct warehouse stock location (`recordset.grouped(...)`), aggregating `reserved_quantity:sum` in SQL.
- No schema/field/view change — pure Python logic.

## Measured (order of 250 lines)
- `stock.quant` queries for these two fields: **~500 → 3** (constant, no longer scales with line count).
- Compute time: **~320 ms → ~10 ms**.
- Output is identical to the previous per-line logic (verified with real stock data: available = on-hand − reserved per location; reserved summed child_of the warehouse stock location).

### DIFF
--- a/sale_stock_ux/models/sale_order_line.py
+++ b/sale_stock_ux/models/sale_order_line.py
@@ -2,6 +2,8 @@
 # For copyright and license notices, see __manifest__.py file in module root
 # directory
 ##############################################################################
+from collections import defaultdict
+
 from odoo import _, api, fields, models
 from odoo.tools.float_utils import float_compare, float_is_zero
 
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1762)

jcadhoc:
@roboadhoc r+ nobump

PR #1763, opened 2026-07-13
[FIX] portal_sale_distributor: add ir.rule to allow distributors to r…

lav-adhoc:
…ead company partners

Portal distributor users were unable to create sale orders for themselves. The sale order form reads company_id.partner_id during rendering and onchanges, but the native portal rule restricts res.partner access to child_of commercial_partner_id, blocking that read with an AccessError.

Add a minimal ir.rule granting the group read-only access to the Odoo company partner(s) the user belongs to — no access to customers, suppliers or any other partner.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1763)

PR #1770, opened 2026-07-20
[FIX] sale_order_type_invoice_policy: detect delivery chain via routing rules when the next leg is not created yet

gal-adhoc:
## Summary

`_is_delivery_chain()` (introduced in #1765) determines whether a picking eventually reaches a customer location by rolling up already-created `stock.move` records via `move_dest_ids`. On routes where the next leg's move is generated lazily — only once the current leg is marked done, instead of upfront at sale confirmation — that rollup finds nothing yet at `button_validate()` time. The prepaid / `prepaid_block_delivery` check in `button_validate()` is then silently skipped on the first leg of any multi-step delivery, which is exactly the leg where stock physically leaves the warehouse.

## Fix

Falls back to walking the warehouse's `stock.rule` graph (`location_src_id` -> `location_dest_id` per active rule) starting from each move's current destination, when the move-based rollup finds no customer location. Rules are static routing configuration and exist regardless of whether the moves they will generate have been created yet.

## Test plan

- Warehouse with a 2+ step delivery route where the next leg's `stock.move` is only created once the current leg is validated (not upfront at confirmation). Confirm a sale order with type `invoice_policy=prepaid`/`prepaid_block_delivery` and an unpaid invoice, then validate the first leg (e.g. Pick): `button_validate()` should now raise the `UserError` instead of going through.
- Regression: standard 3-step delivery (Pick/Pack/Out) where all moves are created upfront still blocks as before (covered by the existing move-based rollup).
- Regression: receipts/returns are not blocked (their rule graph never reaches a customer location).

Ref: ticket 123212.

### DIFF
--- a/sale_order_type_invoice_policy/models/stock_picking.py
+++ b/sale_order_type_invoice_policy/models/stock_picking.py
@@ -72,8 +72,42 @@ class StockPicking(models.Model):
         cover the default 3-step route and miss any extra custom ones.
         Return/receipt pickings are excluded naturally, since their moves
         never end up in a customer location.
+
+        Some routes only create the next leg's stock.move once the current
+        one is done (instead of upfront at sale confirmation), so rolling up
+        move_dest_ids alone misses the chain for pickings validated before
+        that next move exists. Falling back to the warehouse's stock.rule
+        graph covers that case: rules are static routing configuration, so
+        they are already there even when the moves they will generate are
+        not.
+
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1770)

mav-adhoc:
@roboadhoc r+ nobump

PR #1771, opened 2026-07-20
[FIX] sale_stock_ux: refresh stock return picking view on update

les-adhoc:
`sale_stock_ux` overrides the `action_create_returns_all` button, which Odoo added to `stock` within the 18.0 series. `stock` does not bump its manifest on a fix, so its base view may still lack the button in DB and loading our override fails.

This pre-migration refreshes `stock.view_stock_return_picking_form` from XML so the button exists before our view is loaded.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1771)

les-adhoc:
Reemplazado por #1772 (branch renombrado al patrón OBA `18.0-h-123274-les`).

PR #1777, opened 2026-07-21
[FIX] portal_sale_distributor: grant write/create/unlink on sale.orde…

lef-adhoc:
…r.spreadsheet

Portal distributor users could not duplicate a sale order that had a quotation spreadsheet attached: spreadsheet_sale_management.copy() copies the spreadsheet_ids, and spreadsheet.mixin unconditionally runs a write access check on sale.order.spreadsheet. Its _check_access is `super() or order_id._check_access()`, so the model-level ACL must grant the permission first; the group only had read, raising AccessError.

Granting CRUD on the model is safe and scoped: the distributor is subject only to the global multi-company rule, and _check_access still delegates to order_id, keeping access limited to the distributor's own orders.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1777)

matiasperalta1:
@roboadhoc r+ bump

PR #1778, opened 2026-07-22
[ADD] portal_sale_distributor_loyalty: add new module

lef-adhoc:
Bridge module granting the Portal Distributor group read access on sale.order.coupon.points (sale_loyalty). Without it, a portal distributor gets an AccessError when opening a sale order that has loyalty/coupon points, since the group has no ACL on that model.

Kept as a separate auto_install bridge (depends sale_loyalty + portal_sale_distributor) because sale_loyalty is not a dependency of portal_sale_distributor and is not pulled in on its own, so the ACL cannot live in the base module.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1778)

matiasperalta1:
@roboadhoc r+ nobump

PR #1783, opened 2026-07-23
[FIX] sale_gathering: keep manual discount on order confirmation

lef-adhoc:
On gathering orders, _action_confirm moves each line quantity to initial_qty_gathered and writes product_uom_qty = 0. Since discount is a stored computed field depending on product_uom_qty, that write retriggered _compute_discount, which reset the manual discount from the pricelist (e.g. to 0% with a percentage rule).

Protect gathering lines with initial_qty_gathered > 0 from the discount recompute, mirroring the existing _compute_price_unit override.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1783)

matiasperalta1:
@roboadhoc r+ nobump

PR #1785, opened 2026-07-29
[FIX] sale_ux: avoid spurious credit note on fully deducted final invoice

feg-adhoc:
With "Round Globally" tax rounding, invoicing a 100% down payment and then creating the final regular invoice yields a credit note of 0.01 instead of an invoice of 0.00. The sales order is still flagged as fully invoiced, so the customer is left with an unexpected refund document, and cancelling it with a debit note duplicates the quantities.

A down payment line can only store a `price_unit` rounded to the 'Product Price' decimal precision, while the product lines it must offset are aggregated from their raw amounts. When a product subtotal falls on a half cent (e.g. quantity 0.5 at 1.01 => 0.505), the final invoice carries a raw residual of -0.005. `_round_tax_details_base_lines` rounds that aggregate to -0.01 and `_distribute_delta_amount_smoothly` assigns the cent to the largest base line, leaving its `balance` one cent away from its o
`amount_untaxed` derives from the balances, the invoice totals -0.01 and `_create_invoices` switches it to a refund.

The product amounts have already been invoicedent
invoice, so the aggregation must target those rounded amounts. Declare them through `manual_total_excluded_currency`, whicdict and feeds `target_total_excluded`. `total_excluded` is unchanged, so no posted or displayed amount moves.

### DIFF
--- a/sale_ux/models/account_move.py
+++ b/sale_ux/models/account_move.py
@@ -12,6 +12,16 @@ class AccountMove(models.Model):
     sale_order_ids = fields.Many2many("sale.order", compute="_compute_sale_orders")
     # en la ui agregamos este que seria mejor a nivel performance
     has_sales = fields.Boolean(string="Has Sales?", compute="_compute_has_sales")
+    # no almacenado: solo memoiza el chequeo por factura para no recorrer las lineas una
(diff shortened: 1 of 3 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1785)

PR #1788, opened 2026-07-29
[FIX] portal_sale_distributor: allow the distributor group to read warehouses

mav-adhoc:
## Problem

`security/ir.model.access.csv` grants read on `stock.warehouse` to `group_portal_backend_distributor`, but no record rule is declared for that model. A portal user also belongs to `base.group_portal`, and `mrp_subcontracting` declares a rule on that group:

```
[('id', 'in', user.partner_id.commercial_partner_id.picking_ids.picking_type_id.warehouse_id.ids)]
```

While no portal rule existed on `stock.warehouse` the ACL alone was enough. As soon as `mrp_subcontracting` is installed, that rule applies and the distributor only sees the warehouses of its own pickings — a distributor placing its first order has none. Adding a line to a sale order then fails to read the warehouse, the row is never created, and `portal_sale_distributor_tour` times out at "wait for new row to be created".

This is not specific to the tour: any database with `portal_sale_distributor` and `mrp_subcontracting` installed has its distributor users unable to add order lines.

## Fix

A read-only rule on `stock.warehouse` for the distributor group, mirroring `portal_product_template_public_rule` right above it — that one exists for exactly the same reason, since `mrp_subcontracting` also adds a `product.template` rule on `base.group_portal`. Rules of different groups are OR-ed, so this restores the behaviour the module had before `mrp_subcontracting` was in the database, without widening what the ACL already allowed.

## Tests

Verified on a database with `portal_sale_distributor` + `mrp_subcontracting`, using a portal user in the distributor group whose partner has no pickings:

- with the rule: 1 warehouse visible
- with the rule deactivated: 0 warehouses visible, which is the state that breaks the tour

## Context

Surfaced by ingadhoc/product#921, which adds `mrp_subcontracting` to the dependency tree of `product_currency_mrp` and therefore installs it in the OBA build for the first time. Same branch name in both repos so runbot builds them together.

Internal reference: https://www.adhoc.inc/odoo/helpdesk.ticket/122531

### DIFF
--- a/portal_sale_distributor/security/portal_sale_distributor_security.xml
+++ b/portal_sale_distributor/security/portal_sale_distributor_security.xml
@@ -51,15 +51,26 @@
         <field eval="0" name="perm_unlink"/>
     </record>
 
-    <!-- The ir.model.access above grants read on stock.warehouse, but a portal
-         user is also in base.group_portal, and mrp_subcontracting adds a rule
-         on that group restricting warehouses to the ones of the partner's own
-         pickings. A distributor placing its first order has none, so adding a
-         sale order line fails to read the warehouse. Same reason as the
-         product.template rule above. -->
+    <!-- The ir.model.access entries grant read on stock.warehouse and uom.uom,
+         but a portal user is also in base.group_portal, and mrp_subcontracting
+         adds rules on that group restricting both models to what belongs to the
+         partner's own subcontracting (its pickings, its BoMs). A distributor has
+         none of that, so it cannot add a line to a sale order. Same reason as
+         the product.template rule above. -->
     <record id="portal_stock_warehouse_rule" model="ir.rule">
         <field name="name">Warehouses (distribuidores)</field>
         <field name="model_id" ref="stock.model_stock_warehouse"/>
+        <field name="domain_force">[(1 ,'=', 1)]</field>
+        <field name="groups" eval="[(4, ref('group_portal_backend_distributor'))]"/>
+        <field eval="1" name="perm_read"/>
+        <field eval="0" name="perm_write"/>
+        <field eval="0" name="perm_create"/>
+        <field eval="0" name="perm_unlink"/>
+    </record>
+
+    <record id="portal_uom_rule" model="ir.rule">
+        <field name="name">UoM (distribuidores)</field>
+        <field name="model_id" ref="uom.model_uom_uom"/>
         <field name="domain_force">[(1 ,'=', 1)]</field>
         <field name="groups" eval="[(4, ref('group_portal_backend_distributor'))]"/>
         <field eval="1" name="perm_read"/>
--- a/portal_sale_distributor/security/portal_sale_distributor_security.xml
+++ b/portal_sale_distributor/security/portal_sale_distributor_security.xml
@@ -51,4 +51,21 @@
         <field eval="0" name="perm_unlink"/>
     </record>
 
+    <!-- The ir.model.access above grants read on stock.warehouse, but a portal
+         user is also in base.group_portal, and mrp_subcontracting adds a rule
+         on that group restricting warehouses to the ones of the partner's own
+         pickings. A distributor placing its first order has none, so adding a
+         sale order line fails to read the warehouse. Same reason as the
+         product.template rule above. -->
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1788)

mav-adhoc:
## This fix is correct but not sufficient on its own

Verified on the OBA build: with this rule the tour gets past "wait for new row to be created" (step 7/12) and now fails later, at step 11/12, with the same root cause on a different model — `uom.uom`, blocked by `mrp_subcontracting.uom_subcontracting_rule`.

The underlying issue is broader than one model, so I am leaving it here for the module owners to decide rather than patching model by model.

## The pattern

`mrp_subcontracting` declares ~14 record rules on `base.group_portal`, designed for the **subcontractor portal** — a partner that manufactures for us. `portal_sale_distributor` builds on the same base group for a completely different persona: a **distributor placing orders**. Rules of different groups are OR-ed, but a distributor matches none of the subcontracting domains (no BoMs of its own, no subcontracting pickings), so every model covered by those rules becomes unreadable.

Cross-checking the 56 models this module grants ACLs to against those rules, they overlap on 9:

| Model | Status |
|---|---|
| `product.template` | already handled by `portal_product_template_public_rule` |
| `stock.warehouse` | this PR |
| `uom.uom` | currently failing |
| `stock.location`, `stock.picking.type` | pending — configuration data, low risk |
| `stock.picking`, `stock.move`, `stock.move.line`, `stock.lot` | pending — **needs a security decision** |

## Why I stopped

This module currently declares no record rule at all on any stock model — only ACLs. So today, without `mrp_subcontracting` installed, a distributor can already read every picking, move and lot in the database. Adding `[(1,'=',1)]` rules for those four models would restore exactly that behaviour, but it would also be the first time it is written down explicitly, and it is not a call to make from an unrelated bugfix.

It is worth noting this is not a CI-only problem: any database with both modules installed has its distributor users unable to add order lines at all.

Happy to implement whichever direction the owners prefer.

mav-adhoc:
## Scope is smaller than I thought — two rules, none of them sensitive

I reproduced the tour's flow in a shell instead of waiting for runbot, and the earlier estimate of "9 colliding models" was wrong about how many actually matter.

The key detail I had missed: `action_confirm_distributor` calls `self.sudo()` before `action_confirm()`. So everything the confirmation touches — pickings, moves, orderpoints — never goes through the portal user's rights. The failures I was chasing there were artefacts of calling `action_confirm()` directly in my simulation.

What the portal user actually needs to read is only what the form renders before confirming:

| Step | Model | Status |
|---|---|---|
| adding the line | `stock.warehouse` | rule added |
| rendering the line | `uom.uom` | rule added |
| confirming | — | runs as sudo |

`stock.picking`, `stock.move`, `stock.move.line` and `stock.lot` are **not** involved, so no security decision is needed on them. That was my main concern in the previous comment and it does not apply.

## Verification

Simulated the full tour flow on a clean database with `portal_sale_distributor` + `mrp_subcontracting` + `sale_stock`, as a portal user in the distributor group whose partner has no pickings and no BoMs:

```
1. create sale.order                         OK
2. add line                                  OK
3. read line fields                          OK
3b. read the uom.uom record                  OK
4. confirm (action_confirm_distributor)      OK
5. read generated pickings                   OK
```

Deactivating only the `uom.uom` rule reproduces the exact failure the tour hits at step 11/12:

```
3b. read the uom.uom record    ACCESS DENIED
    doesn't have 'read' access to: - Product Unit of Measure (uom.uom)
```

That denial is what raises the error dialog, and the modal is what blocks the click on the confirm button — the tour reports it as "It is not allowed to do action on an element that's below a modal".

les-adhoc:
@roboadhoc r+ bump

roboadhoc:
@mav-adhoc @les-adhoc because this PR has multiple commits, I need to know how to merge it:

* `merge` to merge directly, using the PR as merge commit message
* `rebase-merge` to rebase and merge, using the PR as merge commit message
* `rebase-ff` to rebase and fast-forward

les-adhoc:
@roboadhoc rebase-ff

roboadhoc:
Merge method set to rebase and fast-forward.

PR #1793, opened 2026-07-30
[IMP] sale_order_type_automation: don't confirm the order if the invo…

lef-adhoc:
**Merge order: ingadhoc/account-invoicing#311 goes first.** This PR uses `_notify_background_post_error`, `_get_background_post_error_body` and `_get_background_post_batch_size`, which #311 adds to `account_background_post`. Without it the module degrades to the v18 behaviour (abort with a plain error, no call to action), so nothing breaks, but the feature is only half there.

…ice fails

Until now, when the invoicing automation could not validate the invoice, the error was swallowed: the order was confirmed anyway and the failure only showed up on the chatter. By backend that is undesired -- the user has to see it at once, not end up with an order "confirmed" hiding an error.

Now the whole operation is aborted and a call to action offers to run it again deferring the validation to the background post cron:

* On order confirmation, "Confirm anyway" confirms the order leaving the invoice in draft, marked to be validated in background.
* On transfer validation, "Validate anyway" does the same for the picking.
* Validating more transfers than the background post batch size never tries to validate the invoices synchronously: they are created already deferred.
* Background post errors of these invoices are logged on the source sale order only -- whoever has to act on them follows the order, not the invoice.

The error is logged through a separate database connection, so the trace survives the rollback and it also works when the current cursor was already aborted by a database error.

account_background_post stays optional: without it there is nothing to defer to and the historical behaviour is kept.

Task 70922


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1793)

matiasperalta1:
@roboadhoc r+ bump

PR #1795, opened 2026-07-30
[IMP] sale_order_type_invoice_policy: distinguish MTO receipt block from delivery block

gal-adhoc:
## Context

`button_validate()` blocks picking validation when the related sale order
uses the `prepaid`/`prepaid_block_delivery` invoice policy and hasn't been
fully paid. This correctly covers both:
- Outgoing deliveries to the customer.
- Receipts tied to a purchase generated to fulfill a Make To Order (MTO)
  sale line — the sale being unpaid should also hold up receiving the
  goods that will eventually fulfill it.

Both cases raised the exact same message, which only makes sense for the
delivery case ("...then every sale line must be invoiced and paid before
you can validate picking"). Users validating a receipt had no way to tell
why a purchase-side operation was being blocked by a sale order's invoice
policy.

## Change

- Added a dedicated message for the MTO-receipt case, explaining that the
  receipt exists to fulfill an MTO sale line and that the sale order needs
  to be fully invoiced and paid first.
- Added `_is_customer_delivery()`, a small helper used only to pick which
  of the two messages to show (whether this picking's own moves actually
  end at a customer location, via `stock.move.location_final_id`). The
  blocking condition itself (`_is_delivery_chain()`) is unchanged.
- Added the missing Spanish translations for both messages — they had no
  `i18n` entry at all, so Spanish-language databases were showing them in
  English.

## Test plan

- [ ] Unpaid sale order (MTO, `prepaid_block_delivery` type) generating a
      purchase order:
  - [ ] Validate the incoming receipt → blocked with the new MTO-specific
        message.
  - [ ] In a 2-step reception warehouse, validate the put-away step →
        blocked with the same MTO-specific message.
- [ ] Unpaid sale order, outgoing delivery → still blocked with the
      original delivery message.
- [ ] Both messages render in Spanish on an `es`/`es_419` database.

### DIFF
--- a/sale_order_type_invoice_policy/i18n/es.po
+++ b/sale_order_type_invoice_policy/i18n/es.po
@@ -96,3 +96,30 @@ msgstr "No aplica la política de facturación Cantidades Entregadas para los se
 #: model:ir.model,name:sale_order_type_invoice_policy.model_sale_order_type
 msgid "Type of sale order"
 msgstr "Tipo de pedido de venta"
+
+#. module: sale_order_type_invoice_policy
+#. odoo-python
+#: code:addons/sale_order_type_invoice_policy/models/stock_picking.py:0
+msgid ""
+"If you use a sale type in the sale order related with invoice policy "
+"\"Block Reserve/Block Delivery\", then every sale line must be invoiced "
+"and paid before you can validate picking"
+msgstr ""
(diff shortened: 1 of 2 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1795)

PR #1796, opened 2026-07-31
[IMP] sale_payment_options: discriminate payment options by sale orde…

lef-adhoc:
…r line

New per company setting "Payment Options Display: Discriminate by sale order line", in the Quotations & Orders block of the Sales settings. When it is enabled the printed quotation shows the amount per installment of every payment option under each sale order line, the order wide payment options table is no longer printed, and the standard totals summary is shown again (the module used to hide it unconditionally to replace it with that table).

The installment amounts are computed on the line amount the report already prints, and take the surcharge and the installment quantity from the payment option itself instead of the installment plan, so editing a plan afterwards does not change an already printed quotation. Options split across several plans keep their segments joined and prefixed with their percentage, to avoid printing them as if they were separate alternatives.

Also print the amounts of the payment options table with the monetary widget instead of raw floats, and replace t-esc with t-out.

### DIFF
--- a/sale_payment_options/README.rst
+++ b/sale_payment_options/README.rst
@@ -24,6 +24,7 @@ Allows defining and displaying multiple payment options on quotations and sales 
 - Each payment option can have multiple installment plans.
 - Payment options are shown in the sales order PDF, with tables, subtotals, and totals.
 - Handles missing or malformed data gracefully in reports.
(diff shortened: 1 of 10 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1796)

PR #1587, opened 2026-02-26
[IMP] portal_sale_distributor: add stock group and update product vie…

lef-adhoc:
…ws for better access control

PR #1798, opened 2026-08-03
[FIX] portal_sale_distributor: don't override standard display_stock in product catalog

gal-adhoc:
## Summary

`ProductCatalogMixin._get_action_add_from_catalog_extra_context()` in `portal_sale_distributor` replaced `display_stock` unconditionally based on `group_portal_backend_distributor_stock`, instead of extending the value computed by `super()` (`sale_stock`, `mrp`, etc.). As a result, any user without that specific group lost the on-hand quantity indicator in the product catalog Kanban (`stock.product_view_kanban_catalog`), even for regular internal sale orders with storable products — a regression from standard Odoo behavior.

Introduced in #1587 (@lef-adhoc).

## Fix

Merge into the standard `display_stock` instead of overriding it: the distributor group can only force it `True`, it no longer hides stock for regular backend users.

## Test plan

- As an internal user without `portal_sale_distributor.group_portal_backend_distributor_stock`, open a sale order's product catalog with storable products in stock → "On Hand" shows as in standard Odoo.
- As a user with the distributor group → still shows "On Hand" as before.

Internal ref: Helpdesk ticket #124272.

### DIFF
--- a/portal_sale_distributor/models/product_catalog_mixin.py
+++ b/portal_sale_distributor/models/product_catalog_mixin.py
@@ -5,10 +5,7 @@ class ProductCatalogMixin(models.AbstractModel):
     _inherit = "product.catalog.mixin"
 
     def _get_action_add_from_catalog_extra_context(self):
-        display_stock = (
-            True if self.env.user.has_group("portal_sale_distributor.group_portal_backend_distributor_stock") else False
-        )
-        return {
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1798)

matiasperalta1:
@roboadhoc nobump r+

PR #1801, opened 2026-08-03
[FIX] sale_order_type_automation: fix payment automation on refunds

rov-adhoc:
## Problem

Sale order types with `payment_atomation = validate_payment` created a wrong payment on credit notes:

- `payment_type` was never passed to `account.payment.create()`, so it fell back to the field default (`inbound`). A customer credit note therefore produced a money receipt instead of a refund, the payment could not be reconciled against the credit note, and the resulting accounting entries were reversed.
- `reconciled_invoice_ids` was passed on create, but on `account.payment` that is a non stored computed field with no inverse (`addons/account/models/account_payment.py`), so the write was dropped and the payment was never linked to its invoice.
- The `Payment Journal` domain only required an *inbound* manual payment method. An outbound payment needs an outbound one too, so posting the payment of a credit note can fail with `Please define a payment method line on your payment`.

## Fix

- Derive `partner_type` and `payment_type` from the document itself with `is_sale_document()` / `is_inbound()`. This also covers vendor credit notes (`supplier` + `inbound`).
- Use `invoice_ids`, the stored m2m that actually holds the link, instead of `reconciled_invoice_ids`.
- Require a manual method on both directions in `_compute_payment_journal_domain`, and reword `_check_payment_journal_company`: its message blamed the invoice company even when the real problem was the missing payment method.

## Test plan

New `sale_order_type_automation/tests/test_payment_automation.py`:

| Test | Checks |
|---|---|
| `test_prepare_payment_vals_follow_document_direction` | `partner_type` / `payment_type` for the four invoice move types |
| `test_customer_invoice_registers_inbound_payment` | inbound payment, reconciled, linked to the invoice |
| `test_customer_refund_registers_outbound_payment` | outbound payment, reconciled, linked, receivable line debited |
| `test_payment_journal_requires_manual_method_on_both_directions` | a bank journal without an outbound manual method is rejected |

`test_sale_order_type_automation.py` picks its payment journal with the same domain as the model.

```
odoo -d <db> -u sale_order_type_automation --test-enable --test-tags /sale_order_type_automation
odoo.tests.result: 0 failed, 0 error(s) of 10 tests
```

Every new assertion was checked to fail with the two models reverted (`3 failed, 0 error(s) of 10 tests`), so the tests do cover the regression.

Ref: https://www.adhoc.inc/odoo/helpdesk.ticket/124193

### DIFF
--- a/sale_order_type_automation/models/account_move.py
+++ b/sale_order_type_automation/models/account_move.py
@@ -9,12 +9,16 @@ class AccountMove(models.Model):
     _inherit = "account.move"
 
     def _prepare_dict_account_payment(self, invoice, payment_journal):
-        partner_type = invoice.move_type in ("out_invoice", "out_refund") and "customer" or "supplier"
         return {
-            "reconciled_invoice_ids": [(6, 0, invoice.ids)],
+            # reconciled_invoice_ids is a non stored computed field without inverse, so
+            # writing it here was a no-op. invoice_ids is the stored m2m holding the link
+            "invoice_ids": [(6, 0, invoice.ids)],
             "amount": invoice.amount_residual,
             "partner_id": invoice.partner_id.id,
-            "partner_type": partner_type,
+            "partner_type": "customer" if invoice.is_sale_document() else "supplier",
+            # refunds move the money on the opposite direction of the document they fix,
+            # otherwise a customer refund would be registered as an inbound receipt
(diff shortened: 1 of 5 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1801)

lef-adhoc:
@roboadhoc r+ nobump

PR #1817, opened 2026-08-11
[FIX] sale_loyalty_ux: clear and translatable message when the sale domain does not match

lef-adhoc:
## Problem

`sale_loyalty_ux` rejects a loyalty/promotion program when the sales order does not match the program's sale domain, and it did so with a hardcoded `"SaleOrder not matching"` string. That string is returned as the `error` of `_program_check_compute_points`, so it reaches the user verbatim: the "Enter promo code" dialog in the backend and the discount field in the eCommerce cart. It is in English regardless of the user language, and it says nothing about why the program was rejected or what to do about it.

For comparison, core `sale_loyalty` returns proper translatable messages for the other rejection paths ("This code is invalid (%s).", "A minimum of ... should be purchased ...").

## Changes

- `models/sale_order.py`: the error is now built with `_()` and mentions the program name.
- `models/loyalty_program.py`: new `sale_domain_error_message` field (translatable). The actual reason for the rejection depends on the domain each program configures, so the generic message cannot state it; this lets the program define its own text (e.g. "You must be registered to use this coupon"). Falls back to the generic message when empty.
- `views/loyalty_program_views.xml`: the field is shown next to `sale_domain`, only when a domain is set.
- `i18n/es.po`: Spanish translations for the new terms.

No manifest version bump, no migration needed (new optional field).

## Test plan

On a database with `sale_loyalty_ux` installed:

1. Create a promotion program with a code trigger and a `Sales domain` that the order cannot match.
2. Apply the code on a quotation -> the dialog now shows a full sentence naming the program instead of `SaleOrder not matching`, translated when the user language is Spanish.
3. Set `Sales domain message` on the program -> that text is shown instead.
4. Clear the domain -> the reward is applied as before.
5. Apply a non-existent code -> core message ("This code is invalid (...)") is unchanged.

Steps 1-5 were run through `odoo shell` on an 18.0 database with `en_US` and `es_AR` installed.

### DIFF
--- a/sale_loyalty_ux/i18n/es.po
+++ b/sale_loyalty_ux/i18n/es.po
@@ -26,6 +26,23 @@ msgid "Loyalty Program"
 msgstr "Programa de fidelización"
 
 #. module: sale_loyalty_ux
+#: model:ir.model.fields,field_description:sale_loyalty_ux.field_loyalty_program__not_applicable_message
+#: model_terms:ir.ui.view,arch_db:sale_loyalty_ux.loyalty_program_view_form
+msgid "Message When the Promotion Does Not Apply"
+msgstr "Mensaje cuando no aplica la promoción"
+
+#. module: sale_loyalty_ux
+#: model:ir.model.fields,help:sale_loyalty_ux.field_loyalty_program__not_applicable_message
+msgid ""
+"Message shown to the user when the sales order does not match the sales "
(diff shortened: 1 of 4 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1817)

matiasperalta1:
@roboadhoc r+ bump

PR #1820, opened 2026-08-12
[FIX] sale_loyalty_ux: clear and translatable message when the sale domain does not match

lef-adhoc:
Forward port of #1817 (18.0) to 19.0.

## Problem

`sale_loyalty_ux` rejects a loyalty/promotion program when the sales order does not match the program's sale domain, and it did so with a hardcoded `"SaleOrder not matching"` string. That string is returned as the `error` of `_program_check_compute_points`, so it reaches the user verbatim: the "Enter promo code" dialog in the backend and the discount field in the eCommerce cart. It is in English regardless of the user language, and it says nothing about why the program was rejected or what to do about it.

For comparison, core `sale_loyalty` returns proper translatable messages for the other rejection paths ("This code is invalid (%s).", "A minimum of ... should be purchased ...").

## Changes

- `models/sale_order.py`: the error is now built with `_()` and mentions the program name.
- `models/loyalty_program.py`: new `not_applicable_message` field (translatable). The actual reason for the rejection depends on the domain each program configures, so the generic message cannot state it; this lets the program define its own text (e.g. "You must be registered to use this coupon"). Falls back to the generic message when empty.
- `views/loyalty_program_views.xml`: the field is shown next to `sale_domain`, only when a domain is set.
- `i18n/sale_loyalty_ux.pot` / `i18n/es.po`: new source terms and their Spanish translations.

No migration needed (new optional field).

## Test plan

Run on a clean 19.0 database with `sale_loyalty_ux` installed and `es_AR` activated:

1. Promotion program with a code trigger and a `Sales domain` the order cannot match, applying the code on a quotation:
   - `en_US`: `This promotion cannot be applied because this sales order does not meet the conditions required by the program "Promo Registrados".`
   - `es_AR`: `Esta promoción no puede aplicarse porque el pedido de venta no cumple con las condiciones requeridas por el programa "Promo Registrados".`
2. With `Message When the Promotion Does Not Apply` set on the program, that text is returned instead (per language, since the field is translatable).
3. Clearing the domain applies the reward as before.
4. A non-existent code still returns the core message `This code is invalid (NOEXISTE).`
5. The field renders in the loyalty program form, and its label/help are translated.
6. Module test suite: `4 tests 0.55s 697 queries` — `0 failed, 0 error(s)`.

### DIFF
--- a/sale_loyalty_ux/i18n/es.po
+++ b/sale_loyalty_ux/i18n/es.po
@@ -49,6 +49,23 @@ msgid "Loyalty Rewards Banner Visible"
 msgstr ""
 
 #. module: sale_loyalty_ux
+#: model:ir.model.fields,field_description:sale_loyalty_ux.field_loyalty_program__not_applicable_message
+#: model_terms:ir.ui.view,arch_db:sale_loyalty_ux.loyalty_program_view_form
+msgid "Message When the Promotion Does Not Apply"
+msgstr "Mensaje cuando no aplica la promoción"
+
+#. module: sale_loyalty_ux
+#: model:ir.model.fields,help:sale_loyalty_ux.field_loyalty_program__not_applicable_message
+msgid ""
+"Message shown to the user when the sales order does not match the sales "
+"domain. Use it to explain why the program does not apply and what to do to "
+"make it apply. If empty, a generic message is used."
+msgstr ""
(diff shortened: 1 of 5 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1820)

matiasperalta1:
@roboadhoc r+ bump

PR #1823, opened 2026-08-13
[IMP] sale_barcode: add packaging barcodes and keep scanned quantities whole

lef-adhoc:
`sale_barcode` did not handle packagings. In 19.0 packagings are units of measure (`product.template.uom_ids`, with `product.uom` holding their barcode), and scanning one was rejected as an unknown barcode.

### What changes

- **Packaging barcodes are recognized.** Barcodes are unique across products and packagings, so a barcode matching no product is looked up on `product.uom` — same criterion `stock_barcode` uses. Scanning one adds the product on a line expressed in that packaging unit.
- **Scanned quantities stay whole.** On an existing line the quantity is expressed in the unit of that line: scanning a packaging on a line sold by unit adds its whole content, and scanning a unit on a line sold by packaging adds a whole packaging instead of a fraction of one.
- A packaging not available for sale on the product falls back to the product unit.
- Product lookup now uses `limit=1`; a duplicated barcode used to return a multi-record set.

### Test plan

`sale_barcode/tests/test_sale_barcode.py` drives the real `_barcode_scanned` onchange through `Form`. On a product sold by unit with a `Pack of 12` packaging:

| Scans | Result |
|---|---|
| 15 × unit | `15.00 Units` |
| 3 × packaging | `3.00 Pack of 12` |
| 2 × unit, then packaging | `14.00 Units` |
| packaging, then unit | `2.00 Pack of 12` |
| unit, packaging set by hand on the line, then unit | `2.00 Pack of 12` |
| packaging, with the packaging removed from the product | `12.00 Units` |
| unknown barcode | warning, no line |

Also checked in the browser on a fresh 19.0 database: quantities and the packaging unit hold after saving, and the unit price converts per unit of measure.

Task: /odoo/project.task/72469

PR #1824, opened 2026-08-14
[IMP] sale_barcode: document the packaging circuit and its known limitation

lef-adhoc:
Documents the packaging circuit in the module README, so the behaviour and the case we analysed and chose not to cover are written down where the module lives.

### What it documents

- Where a packaging lives in 19.0 (a unit of measure on the product's *Sales* tab, with its barcode under *Packaging Barcodes*) and what each barcode does when scanned on a sale order.
- That the quantity of an existing line is always expressed in the unit of that line.

### The known limitation

Scanning a packaging of 6 units and then the same product by unit leaves the line at 2 packagings — 12 units — instead of the 7 that were scanned, because the scan is added in the unit of the existing line.

Adding the exact amount instead would put a fraction of a packaging on the line, and a decimal quantity where the salesperson counted whole items. Selling a packaging plus some loose units of the same product therefore needs the loose units set by hand on their own line.

Kept as is for now rather than splitting a line per unit of measure, since we do not expect it to be a common case.

### Notes

Documentation only, no code. Complements #1823, which adds the packaging barcode handling described here.

Task: /odoo/project.task/72469

### DIFF
--- a/sale_barcode/README.rst
+++ b/sale_barcode/README.rst
@@ -39,6 +39,40 @@ To use this module, you need to:
 
 #. To view Sale Orders Lines, go to: Sales/Sales/Sale Lines
 
+Packagings
+----------
+
+A packaging is a unit of measure of the product, listed on its *Sales* tab
+under *Packagings*, and its barcode is set on that unit of measure (*Packaging
+Barcodes*). Both barcodes are taken into account when scanning on a sale order:
+
+* Scanning the product barcode adds one unit.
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1824)

gal-adhoc:
@roboadhoc r+ nobump

PR #1826, opened 2026-08-18
[FIX] sale_ux: keep the final invoice balanced when a down payment is deducted

gal-adhoc:
### What

With "Round Globally" tax rounding, invoicing a down payment and then creating the final invoice can leave the final invoice one cent away from the sales order total. When the down payment covers the order entirely, the final invoice even comes out as a **credit note of 0.01** instead of an invoice of 0.00 — and cancelling that credit note with a debit note duplicates the quantities.

### Why it happens

The down payment wizard rounds the down payment `price_unit` to the currency **once per (taxes, account) combination** (`sale/wizard/sale_make_invoice_advance.py`, `_prepare_down_payment_lines_values`). The product lines it offsets are aggregated from their raw amounts, so when a subtotal falls below the currency precision — a discount such as 15% off 1000.10 => 850.085, or a decimal quantity such as 0.5 at 1.01 => 0.505 — the residual of the whole set is amplified by `round_globally` into a full cent. That cent lands on a base line and can flip the sign of the total.

### Fix

Aggregate and round the product lines of the final invoice with the same partition the down payment used, and feed the result through `manual_total_excluded_currency`. `price_unit`, `price_subtotal` and the raw amounts are untouched; what moves is the line's `total_excluded` — and with it its accounting balance — by the cent that is precisely the bug.

Each down payment line of the invoice already carries the key of its own group, so the partition is **read from those lines** — by order and taxes — rather than derived again. The wizard also groups by account, but the account that ends up on the down payment line cannot be re-derived reliably from the product, so rather than guessing it the code checks separately that the down payment lines of a group share a single one, and leaves the group alone when they do not. A group can hold more than one down payment line — that happens when an order was paid in advance in several rounds — and the target is the same either way: the aggregate of the group rounded once. A group of product lines with no down payment line of its own is left alone, so an unforeseen setup falls back to the standard rounding instead of risking a different discrepancy.

The adjustment is applied **only through `_get_rounded_base_and_tax_lines`**, the path that produces the accounting amounts and the invoice totals. Other consumers of the per-line hook keep seeing the raw base — the early payment discount scales `price_unit` after building the base line, and several EDIs aggregate a subset of the lines, so a pinned group base would give them a wrong amount.

The per-invoice map is memoized in a non-stored computed field, so the hook — which core calls once per line — stays O(n) on the number of invoice lines.

Restricted to `round_globally`, the only rounding method where this residual appears.

### Test plan

`sale_ux/tests/test_downpayment_rounding.py`, all under `round_globally`:

| Scenario | Expected |
|---|---|
| 1 line qty 0.5 @ 1.01, 100% down payment | final invoice of 0.00, not a credit note |
| 3 lines, 100% | final 0.00; down payment + final == order total |
| 5 lines, 100% | idem — the error used to grow with the number of lines |
| 3 lines, 40% | down payment + final == order total |
| 2 different taxes, 100% | two down payment lines, final 0.00 |
| 3 lines without residual, 30% | unchanged (90.00 + 210.00 on 300.00) |
| a down payment account configured on the product category | final 0.00 — the partition must not depend on re-deriving that account |
| residual from a **discount**, down payment charged in **two rounds** | final 0.00, not a credit note; the three documents sum to the order total |
| a fixed tax on some lines only, 50% | invoiced base == order base (the group must not be split) |
| consolidated final invoice deducting one down payment per order | final 0.00, not a credit note |
| `round_per_line` | adjustment stays inactive |

### Scope

Core handles this natively from 18.3 onwards — the tax engine gained dedicated down payment support (`account.tax._prepare_down_payment_lines` / `_prepare_base_lines_for_down_payment`, verified on 19.0) and sets the same `manual_total_excluded_currency` this override sets. So this is meant for **18.0 only and must not be forward-ported to 19.0**, where it would overwrite the value core computes. Suggested merge: `fw=no`.

### Known limitations

- The **down payment itself** can end one cent above the order total, and then the final invoice is negative by that cent. Two ways in: several tax groups (the wizard rounds per group while the order aggregates everything at once), or a down payment charged in several rounds (each round rounds its own amount, so 2 x 50% of 1.01 bills 0.51 + 0.51). Both are the wizard's rounding, happen with or without this change — where there is no sub-cent residual the adjustment is a no-op — and are out of scope here. `test_downpayment_rounded_up_in_several_rounds` pins the behaviour; what holds either way is that the invoiced total still matches the order total.
- Invoices **posted before** this change keep their stored amounts, while the tax totals widget recomputes on every render, so such an invoice can display a total one cent away from what is booked. It only affects invoices that already carry the defect; they need the usual accounting correction, not a recompute.
- `_round_tax_details_base_lines` is mirrored in `account_tax.js`, so while the final invoice is still a draft the browser widget can show a total that differs until it is saved. In practice the final invoice is created by `_create_invoices` rather than typed by hand.

### DIFF
--- a/sale_ux/models/account_move.py
+++ b/sale_ux/models/account_move.py
@@ -2,6 +2,8 @@
 # For copyright and license notices, see __manifest__.py file in module root
 # directory
 ##############################################################################
+from collections import defaultdict
+
 from odoo import api, fields, models
 
 
@@ -12,6 +14,10 @@ class AccountMove(models.Model):
     sale_order_ids = fields.Many2many("sale.order", compute="_compute_sale_orders")
     # en la ui agregamos este que seria mejor a nivel performance
     has_sales = fields.Boolean(string="Has Sales?", compute="_compute_has_sales")
+    # no almacenado: memoiza el calculo por factura, porque
+    # _prepare_product_base_line_for_taxes_computation corre una vez por linea y necesita el
+    # agregado de todas
+    downpayment_base_targets = fields.Json(compute="_compute_downpayment_base_targets", exportable=False)
 
     @api.depends("move_type", "partner_id", "partner_id.lang", "company_id")
     def _compute_narration(self):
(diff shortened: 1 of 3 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1826)

feg-adhoc (commented):
Revisé el fix (round_globally + anticipos). El enfoque está bien y los tests cubren buenos escenarios. Van tres puntos, uno bloqueante.

**[BLOQUEANTE] El ajuste se activa también en facturas de compra** — `sale_ux/models/account_move.py:82`

`is_invoice(include_receipts=True)` admite `in_invoice`/`in_refund`, y `purchase` también define `is_downpayment` en `account.move.line` (`purchase/models/account_invoice.py:521`), seteándolo en los bills de anticipo de compra. En un bill de anticipo bajo `round_globally`, `_get_downpayment_group_key` agrupa por `sale_line_ids.order_id` (vacío en compras → `()`), juntando líneas de distintas OC en un solo grupo → target equivocado → reintroduce el descuadre, en un módulo que solo debería tocar ventas.

→ Acotar el guard a `move.is_sale_document(include_receipts=True)` en vez de `is_invoice(...)`.

**[MENOR] El anticipo cobrado en varias tandas puede no cerrar** — `account_move.py:115`

El target es `round(Σ raw)` del grupo, pero con anticipo en tandas la deducción real es `Σ round(Σ·ratio)`, que no siempre iguala `round(Σ)` (grupo chico `Σ=0.03` → `2·round(0.015)=0.04` vs `round(0.03)=0.03` → final −0.01 → NC). `test_full_downpayment_in_two_rounds_with_discount` cierra por los montos elegidos, no en el caso general.

→ Caso muy de borde. Sugerencia: documentarlo en "Known limitations" y no bloquear el merge por esto. Si se quiere cubrir, apuntar el target a la deducción real (el `price_subtotal` de las líneas de anticipo del grupo) en vez de recalcular `round(Σ raw)` — de paso baja el acoplamiento a la partición del wizard.

**[MENOR] Falta un depends** — `account_move.py:49`

El campo gatea todo en `company_id.tax_calculation_rounding_method`, pero ese campo no está en el `@api.depends`. Conviene declararlo para que no quede un valor stale si cambia el método de redondeo sin invalidar caché.

Con el bloqueante resuelto (es una línea) queda listo para mergear.

feg-adhoc:
@roboadhoc r+ nobump

PR #1830, opened 2026-08-24
[IMP] sale_payment_options: discriminate payment options by sale order line

gal-adhoc:
Alternative to #1796, built on @lef-adhoc's commit (authorship kept). Same
feature, plus two report fixes found while validating it. Opened as a separate PR
so that #1796 stays untouched and its author can decide which one to move forward
with.

## What comes from #1796, unchanged

- Per company setting "Payment Options Display: Discriminate by sale order line".
- Installment amounts printed under each sale order line, computed on the line
  amount taxes included.
- Surcharge and installment quantity read from the payment option itself, so
  editing a plan afterwards does not change an already printed quotation.
- Options split across several plans keep their segments joined.
- Payment options table amounts printed with the monetary widget, `t-esc` to
  `t-out`.

## What this PR adds

**Drop the report action of the module and its wrapper template.**
`report_saleorder_payment_options` is a primary template that t-calls an
*extension* of `sale.report_saleorder_document`, so it renders the combined arch
of the base document and never reaches the primary templates that localizations
register on it. With `l10n_ar_sale` installed, printing from that entry ignores
the tax discrimination of the sale order type: it prints net unit prices and a
discriminated tax summary on documents that must not discriminate taxes, while
the standard quotation report prints the very same order correctly. It is
redundant on top of that, because the payment options are added by extending
`sale.report_saleorder_document`, so they already print on the standard quotation
report.

**Hide the standard totals summary only when the order wide payment options table
is actually printed.** It used to be hidden unconditionally, so installing the
module removed the totals from every quotation, even from orders without any
payment option.

**No manifest version bump**, left to the mergebot.

## Test plan

On a database with `l10n_ar_sale` and `l10n_ar_sale_order_type`:

1. Sale order type with *Discriminate Taxes* = `No`, and a customer whose tax
   responsibility does not discriminate.
2. Quotation with a couple of lines and several payment options.
3. Setting off: the standard quotation report prints the payment options table
   and no totals summary. A quotation *without* payment options keeps its totals
   summary (that is the second fix).
4. Setting on: the standard quotation report prints the installments under each
   line plus the totals summary, with unit prices and taxes as the localization
   defines them.
5. Installment amounts match `line.price_total * percentage * (1 + surcharge) /
   installments`, and do not change when the order type discriminates taxes.

Not covered here: no automated tests yet, and the rounding of the printed
installment is left as is, so the printed installments can add up to a cent apart
from the line amount.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1830)

matiasperalta1:
@roboadhoc r+ nobump

PR #1631, opened 2026-04-16
[IMP] sale_order_type_automation: add sale order and invoice exclusio…

matiasperalta1:
…n for auto-validation

PR #1831, opened 2026-08-24
[FIX] sale_order_type_automation: invert invoice_validate_domain on upgrade

nicomacr:
## What

`invoice_validate_domain` on `sale.order.type` changed meaning in `19.0.2.1.0` (cf5ea18, #1631):

- **before**: `invoices.filtered_domain(domain)` — matching invoices are the ones that get **validated**
- **now**: `invoices - invoices.filtered_domain(domain)` — matching invoices are the ones left **in draft**

Stored values still carry the old meaning, so after the upgrade automatic invoice validation works exactly the other way around, with no error and no warning. `sale_gathering_automation` consumes the same field and got the same flip, so one fix covers both.

This adds the data migration that was missing: a post-migration that negates every stored domain.

## How

- **Literal domains** go through `~Domain(...)`, which normalizes the implicit AND and applies De Morgan, so the value stays readable in the `domain` widget — `[('move_type', '=', 'out_invoice')]` becomes `[('move_type', '!=', 'out_invoice')]`.
- **Domains holding dynamic expressions** are rewritten **textually**, keeping each element's source verbatim. The field is rendered with `allow_expressions: True`, so it may contain `context_today()` or `relativedelta(...)`; evaluating them here would freeze the expression into the migration date.
- **Empty domains are skipped.** They mean "validate everything" in both versions, and negating one yields a domain that matches nothing.
- Anything that cannot be parsed is logged as a warning and left untouched for manual review.

Note that wrapping the domain in `['!', ...]` is not enough: `!` is unary, so prefixing a two-condition domain yields `NOT c1 AND c2`. It happens to be correct for single-condition domains, which is why a naive version passes a superficial test.

## Why `19.0.2.1.0` and not `19.0.0.0`

That is the version the change shipped in. Pinning it there also covers databases already sitting on `19.0.2.0.0`, which are affected too.

## Test plan

Negation was verified case by case:

- `[c1]` — De Morgan of the condition
- `[c1, c2]` (implicit AND) — `'|'` of both negated conditions
- `['&', c1, c2]` and `['|', c1, c2]` — De Morgan, operator swapped
- `[c1, '|', c2, c3]` — nested, operators swapped at both levels
- `['!', c1]` — double negation removed
- `[]`, and anything unparseable — left untouched
- `[('invoice_date', '>=', context_today())]` — negated with the expression kept verbatim, not evaluated

To check on a build: set a domain on a sale order type before upgrading, then confirm afterwards that the migrated domain selects exactly the complement of what the original one selected, and that re-running `-u` does not invert it a second time.

Task: https://www.adhoc.inc/odoo/project.task/72535

### DIFF
--- a/sale_order_type_automation/migrations/19.0.2.1.0/post-migration.py
+++ b/sale_order_type_automation/migrations/19.0.2.1.0/post-migration.py
@@ -1,0 +1,133 @@
+import ast
+import logging
+
+from odoo.fields import Domain
+
+_logger = logging.getLogger(__name__)
+
+# Arity of the domain prefix operators, used to walk a domain without evaluating it.
+OPERATORS = {"!": 1, "&": 2, "|": 2}
+
+
+def _negate_literal(raw):
+    """Negate a domain whose every element is a literal.
+
+    Returns the negated domain as a string, or None if `raw` is not a plain
+    literal domain (it may contain dynamic expressions, see `_negate_source`).
+    """
+    parsed = ast.literal_eval(raw)
+    if not isinstance(parsed, (list, tuple)) or not parsed:
+        return None
+    # Domain normalizes the implicit AND and applies De Morgan, so the stored
+    # value stays readable in the `domain` widget.
+    return str(~Domain(list(parsed)))
+
+
+def _negate_source(raw):
+    """Negate a domain textually, keeping every element's source verbatim.
+
+    Used when the domain holds dynamic expressions (the field is rendered with
+    ``allow_expressions: True``, so it may contain ``context_today()`` or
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1831)

nicomacr:
Se mueve a `odoo-upgrade` (ingadhoc/odoo-upgrade#203): la corrección tiene que correr solo durante la actualización de versión, para no alterar las bases que ya migraron — un script in-module correría también en un `-u` de mantenimiento y podría re-invertir un dominio ya corregido a mano.

PR #1755, opened 2026-07-03
[FIX] sale_order_type_automation: revert savepoint

lef-adhoc:

PR #1832, opened 2026-08-25
[FIX] sale_order_type_automation: block confirmation on invalid invoice

cem-adhoc:
[FIX] sale_order_type_automation: block confirmation on invalid invoice

- Add account.move._get_invoicing_automation_blocking_reasons: reasons why an
  invoice cannot be posted (latam journal requiring a document type with none
  determined, hinting at the customer's missing ARCA responsibility). Soft
  dependency, no-op without the latam localization
- Raise UserError before posting when there is a reason, restoring the 18.0
  behaviour dropped in #1688 and not put back by the #1755 revert. Batch runs
  (commit_invoice_automation) already committed: they skip the post and log
- Add account.move._recover_failed_automatic_post, called from the except: a
  failed _post leaves state='posted' dirty in cache and the failure log itself
  flushes it, persisting an invoice posted with no number nor document type
- Add tests for the three paths, one reproducing the corrupt state

Change note: Al confirmar un pedido de venta con automatización de facturas, si la factura no puede emitirse porque falta un dato — por ejemplo, la Responsabilidad ARCA del contacto — el sistema vuelve a frenar la confirmación indicando qué corregir, como en la versión anterior. Antes el pedido se confirmaba igual y quedaba una factura registrada sin número ni tipo de documento.

### DIFF
--- a/sale_order_type_automation/models/account_move.py
+++ b/sale_order_type_automation/models/account_move.py
@@ -2,11 +2,58 @@
 # For copyright and license notices, see __manifest__.py file in module root
 # directory
 ##############################################################################
-from odoo import fields, models
+from odoo import _, fields, models
 
 
 class AccountMove(models.Model):
     _inherit = "account.move"
+
+    def _get_invoicing_automation_blocking_reasons(self):
+        """Return ``{move.id: reason}`` for invoices that must not be posted automatically.
+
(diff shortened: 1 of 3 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1832)

PR #1834, opened 2026-08-26
[FIX] sale_ux: run cancellation checks on the mass cancel wizard

gal-adhoc:
The standard `sale.mass.cancel.orders` wizard cancels the selected orders with `_action_cancel()` (`addons/sale/wizard/mass_cancel_orders.py`), which does not run the checks we add on `action_cancel()`:

- `sale_ux`: related invoices in a state other than draft/cancel.
- `sale_stock_ux`: deliveries already in done.

So the "Cancel" action of the sales orders list view cancelled confirmed orders that had posted invoices and validated pickings, leaving those documents untouched and the data inconsistent. The wizard form only shows a warning ("Some confirmed orders are selected…") and cancels anyway.

## Approach

The checks move into `_check_cancel_allowed()`, called both by `action_cancel()` and by the wizard, so the individual and the mass path behave the same:

```python
def action_mass_cancel(self):
    self.sale_order_ids.filtered(lambda order: order.state != "cancel")._check_cancel_allowed()
    return super().action_mass_cancel()
```

`sale_stock_ux` extends that method with its done-deliveries check instead of raising from its own `action_cancel()`, so a single place declares what blocks a cancellation.

The invoice check lives in `_get_orders_blocking_cancel()`, which resolves the whole recordset with one `account.move.line` search and prefetches the fields it reads, so it costs a fixed number of queries regardless of how many orders are selected. `action_cancel()` is multi record safe as a result — it used to raise a singleton error on `inv.invoice_origin == self.name`.

Measured on this branch:

| | Queries |
|---|---|
| Cancelling 200 quotations from the list view | 16 (211 before) |
| Same call, standard wizard with no checks at all | 10 |
| Blocking 10 invoiced orders | 8 |
| Blocking 40 invoiced orders | 8 |

Two behaviour changes that come with it:

- The error names every blocking order instead of failing on the first one. From the list view the user only sees their selection, so a message without names left them with no way to find the culprit.
- Orders already in `cancel` are skipped, so the ones the old path cancelled despite an invoice or a delivery do not block a whole batch.

## Test plan

New tests, all failing on 19.0 without this commit:

- `sale_ux`: `test_mass_cancel_blocks_when_related_invoices_exist`, `test_mass_cancel_allows_when_invoices_are_draft`, `test_mass_cancel_cancels_several_quotations`, `test_mass_cancel_error_names_the_blocking_order`, `test_mass_cancel_skips_already_cancelled_orders`, `test_action_cancel_on_several_orders_names_every_blocking_one`.
- `sale_stock_ux`: `test_mass_cancel_blocks_when_deliveries_are_done`.

Full `TestSaleOrder` + `TestActionCancel` run green (29 tests), so the extracted checks keep the previous semantics.

Manual check: confirm an order, validate its delivery, post its invoice, then select it in the sales list view and use Action > Cancel. The cancellation is now blocked, naming the order.

Note on `sale_stock_ux/tests/test_action_cancel.py`: the shared `setUp` calls `picking._action_done()` without setting quantities or `picked`, so the picking stays in `assigned` and the done-deliveries check was never actually exercised. The new test builds its own done picking instead of relying on it. Fixing the `setUp` is left out of this PR.

Internal reference: https://www.adhoc.inc/odoo/helpdesk.ticket/126450

### DIFF
--- a/sale_ux/wizards/mass_cancel_orders.py
+++ b/sale_ux/wizards/mass_cancel_orders.py
@@ -1,0 +1,16 @@
+##############################################################################
+# For copyright and license notices, see __manifest__.py file in module root
+# directory
+##############################################################################
+from odoo import models
+
+
+class SaleMassCancelOrders(models.TransientModel):
+    _inherit = "sale.mass.cancel.orders"
+
+    def action_mass_cancel(self):
+        """The standard wizard cancels without the checks action_cancel() runs, so we
+        run them here too."""
+        # cancelled orders are skipped: the old path let some through with an invoice
+        self.sale_order_ids.filtered(lambda order: order.state != "cancel")._check_cancel_allowed()
+        return super().action_mass_cancel()
--- a/sale_stock_ux/models/sale_order.py
+++ b/sale_stock_ux/models/sale_order.py
@@ -35,13 +35,19 @@ class SaleOrder(models.Model):
         for order in self:
             order.with_returns = any(line.quantity_returned for line in order.order_line)
 
-    def action_cancel(self):
-        self = self.with_context(cancel_from_order=True)
-        for order in self.filtered(lambda order: order.picking_ids.filtered(lambda x: x.state == "done")):
+    def _check_cancel_allowed(self):
+        delivered = self.filtered(lambda order: order.picking_ids.filtered(lambda x: x.state == "done"))
+        if delivered:
             raise UserError(
-                _("Unable to cancel sale order %s as some deliveries" " have already been done.") % (order.name)
+                _(
+                    "Unable to cancel sale order %s as some deliveries have already been done.",
+                    ", ".join(delivered.mapped("display_name")),
+                )
             )
-        return super().action_cancel()
+        return super()._check_cancel_allowed()
+
+    def action_cancel(self):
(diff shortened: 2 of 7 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1834)

sale_ux/wizards/mass_cancel_orders.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class SaleMassCancelOrders(models.TransientModel):
+    _inherit = "sale.mass.cancel.orders"
+
+    def action_mass_cancel(self):
+        """The standard wizard calls _action_cancel(), skipping the checks we add on
+        action_cancel()."""
+        # cancelled orders are skipped: the old path let some through with an invoice
+        return self.sale_order_ids.filtered(lambda order: order.state != "cancel").action_cancel()
```
lef-adhoc: Aca habría que llamar a "_get_orders_blocking_cancel" y luego return super().action_mass_cancel()

lef-adhoc (changes requested):

lef-adhoc:
@roboadhoc r+ nobump

PR #1835, opened 2026-08-26
[ADD] sale_ux: invariants battery for the old quotations cron

augusto-weiss:
Adds tests/invariants.py with SaleUxInvariants, a mixin the suites pull in through SaleUxCommon, plus its own tests in both directions.

assert_cron_cancelled_exactly checks that exactly the expected quotations ended up cancelled and that every other order in the universe kept the state it had before the operation. The check runs over the records the test created itself, so unrelated orders living in the database can never relax it.

The collateral-damage half is the point: a suite that only asserts "the one I expected got cancelled" stays green while the record next to it breaks.

test_invariants.py verifies the battery in both senses -- it stays quiet on a sane cron run, and it catches an extra cancellation, a missing one and a collateral state change.

### DIFF
--- a/sale_ux/tests/__init__.py
+++ b/sale_ux/tests/__init__.py
@@ -7,6 +7,7 @@ from . import test_account_move
 from . import test_partner_product_config
 from . import test_sale_order
 from . import test_sale_order_line
+from . import test_invariants
 from . import test_settings
 from . import test_wizards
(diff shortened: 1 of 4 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1835)

PR #1841, opened 2026-09-01
[IMP] sale_require_purchase_order_number: send PO number to Chilean DTE

gal-adhoc:
Task: https://www.adhoc.inc/odoo/project.task/70070

## Context

This module stores and validates the customer purchase order number, but it does not
print it or report it anywhere: that is done by each localization. Today only two do it:

| Document | Module |
|---|---|
| AR invoice and delivery slip PDF | `l10n_ar_ux`, `l10n_ar_stock_ux` |
| UY CFE (`CompraID` tag) | `l10n_uy_ux` |
| **CL** | **nothing** |

In Chile the customer PO travels as an ODC cross reference (`l10n_cl.edi.reference`,
doc type `l10n_cl.dc_odc`), which reaches both the DTE and the references table of the
PDF. Two gaps:

1. `l10n_cl_edi_stock` builds the delivery guide reference from `client_order_ref`, and
   returns early when that field is empty — so filling only `purchase_order_number`
   produced no reference at all and the PO never reached SII.
2. On invoices, `l10n_cl_edi` fills references only by hand or from an incoming DTE, so
   nothing was created when invoicing a sale order.

## Changes

- `stock.picking._set_l10n_cl_purchase_order_reference()`, called from
  `sale.order.action_confirm()`: corrects the ODC reference to use
  `purchase_order_number`, or creates it when upstream built none.
- `account.move._set_l10n_cl_purchase_order_reference()`, called from
  `sale.order._create_invoices()`.
- README: new "Regional scope" section stating what each localization does. The
  manifest `summary` was empty; filled with the same idea.

Both guarded on `l10n_cl_reference_ids` being present, `account_fiscal_country_id ==
'CL'` and a configured `l10n_cl_dte_service_provider`, so nothing runs elsewhere. A
reference the user edited is never overwritten, and backorders keep the copy
`l10n_cl_edi_stock` makes.

### Why not in `_get_new_picking_values()`

`l10n_cl_edi_stock` sits above this module in the MRO (it depends on `l10n_cl_edi`,
this module only on `sale_stock`), so it runs first and overwrites
`values['l10n_cl_reference_ids']` after our `super()` call returns. Measured on a
real database:

```
l10n_cl_edi_stock.models.stock_picking        <- more derived
l10n_cl_edi.models.l10n_cl_edi_util
sale_require_purchase_order_number...
```

`stock.picking.create()` does not work either: `sale_id` is not resolved yet there, and
`purchase_order_number` is a non-stored compute that depends on it.

## Test plan

On a database with `l10n_cl`, `l10n_cl_edi`, `l10n_cl_edi_stock` and this module, on a
company with CL fiscal country and a DTE service provider set:

1. SO with `purchase_order_number` and **empty** "Your Reference" → confirm. The
   delivery has one ODC reference whose origin document number is the PO number
   (before: no reference at all).
2. SO with both fields → confirm. One ODC reference, holding the PO number, not
   `client_order_ref`.
3. Edit that reference by hand and re-run the method → the typed value is kept.
4. Invoice case 1 → the invoice carries the PO number and one ODC reference with it.
5. Re-run the method on the invoice → no duplicate.
6. Switch the company to AR and repeat → no Chilean reference is created, the field is
   still stored.

All six verified on Odoo 19.0 (14 assertions).

### DIFF
--- a/sale_require_purchase_order_number/README.rst
+++ b/sale_require_purchase_order_number/README.rst
@@ -20,6 +20,24 @@ This module incorporates the following features:
 * Validate that the field is in these documents if the partner has the field "required number of PO".
 * Purchase order number must be unique per partner
 
+Regional scope
+==============
+
+This module only stores and validates the field. Printing it on a document, or
+sending it to a tax authority, is done by each localization, so what you get out
+of the box depends on the country:
+
+* **Argentina** (``l10n_ar_ux``, ``l10n_ar_stock_ux``): printed as "PO Number" on
+  the invoice and on the delivery slip PDF.
+* **Uruguay** (``l10n_uy_ux``): sent to DGI in the ``CompraID`` tag of the CFE,
+  replacing the value taken from ``ref``.
+* **Chile** (``l10n_cl_edi``, ``l10n_cl_edi_stock``): sent to SII as an ODC cross
+  reference of the invoice and of the delivery guide, and printed on both PDFs in
+  the references table. Without this module that cross reference is built from
+  "Your Reference" (``client_order_ref``) instead.
+* **Any other country**: the field is only stored on the record, it is not printed
+  on any document.
+
 Installation
 ============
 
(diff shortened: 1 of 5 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1841)

PR #1842, opened 2026-09-04
[FIX] sale_gathering: detect the gathering invoice cross-company

lef-adhoc:
## Problem

`sale.order.has_gathering_invoice` is computed from `invoice_ids`, which core
derives from `order_line.invoice_lines`. The multi-company record rule on
`account.move.line` filters that relation, so when the gathering invoice belongs
to a company the user has disabled in the company selector the relation comes
back **empty** and an already invoiced order looks uninvoiced.

Two visible symptoms, one root cause:

| Where | Outcome |
|---|---|
| `sale.order` form | The "Create Gathering Invoice" button shows up again on an order that already has its down payment. Using it adds a **second down payment line and a second invoice** to the order. |
| `sale.order.line._check_gathering_invoice` | Legitimate withdrawals are rejected with *"before adding quantities, you need to create, confirm and pay the gathering invoice"*. |

Both flip back to correct behaviour as soon as the user enables the invoice's
company, which is what makes it look intermittent.

Same root cause as #1833 (`gathering_balance`), through the other two readers of
the relation.

## Fix

Resolve the down payment with a grouped query on `account.move.line`, factored
into `sale.order._get_orders_with_gathering_invoice()`, instead of reading the
relation at all.

Invalidating the cached relation and re-reading it through `sudo()` — the first
approach taken here — returns the right answer but writes the unfiltered records
into the very cache entry the user's own reads go through. `invoice_count` is
computed before `has_gathering_invoice` and keeps the filtered value, so the two
end up disagreeing and the invoice smart button reports a count that does not
match the records the order then hands back; reading the order afterwards can
raise an `AccessError`. A query answers the same question and writes nothing to
that cache.

The domain replicates `account.move._is_downpayment()` (the move must carry no
regular sale line), so an order invoiced with *factura en cero descontando
acopio* is not mistaken for one that already has its down payment.

The wizard is guarded too: a gathering order takes exactly one down payment, so
reaching `create_invoices` with `fixed` or `percentage` on an order that already
has its gathering invoice now raises instead of silently duplicating it. This
closes the double down payment path regardless of which companies the user has
enabled, and no longer depends on the button being hidden.

## Test plan

Verified on a restored copy of a customer database with two companies, where the
gathering order belongs to company A (HIERROSAN) and its down payment invoice to
company B (RAMON SANCHEZ). A real (non-superuser) user, cache invalidated before
each read to emulate a fresh request.

1. Smart button coherence on the order from the report, read the way the web
   client reads a form (all fields at once):

   | | both companies enabled | company B disabled |
   |---|---|---|
   | **Before** (`sudo()` + invalidation) | `invoice_count=3`, `invoice_ids`=3 | `invoice_count=0`, `invoice_ids`=3, action domain lists 3 records the user cannot read |
   | **After** | `invoice_count=3`, `invoice_ids`=3 | `invoice_count=0`, `invoice_ids`=0 |

   Before the change the smart button also depended on the order the fields were
   requested in: asking for `has_gathering_invoice` first yielded
   `invoice_count=3` instead of `0`.

2. `has_gathering_invoice` is `True` in both contexts, so the button stays
   hidden, and `_check_gathering_invoice` lets the withdrawal through in both.

3. Regression over **500** gathering orders in state `sale`, against the original
   pre-fix logic evaluated with both companies enabled as ground truth:
   **0 differences** with `allowed_company_ids=[A, B]` and **0 differences** with
   `[A]` only. The 2 orders with no down payment line still report `False`, so
   the button keeps showing where it should.

4. The wizard guard rejects a second down payment on that order in both company
   contexts.

Task: https://www.adhoc.inc/odoo/helpdesk.ticket/126227


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1842)

gal-adhoc:
@roboadhoc r+ nobump

PR #1846, opened 2026-09-08
[ADD] sale_stock_ux: warn on phantom counter-deliveries from cancel remaining

jcadhoc:
## What

When cancelling the remaining qty of a sale order line on a multi-step delivery route, core lowers `product_uom_qty` and launches a negative procurement (the "mirror"). If that mirror does not merge back into the original outgoing move (the merge key differs by `location_id`, `date_deadline`, `price_unit`, ...), it is not netted, gets reversed and survives as a **phantom counter-delivery** (Customer -> internal), leaving the original move orphaned. In multi-step it is a whole reverse chain (Customer->Despacho + internal put-back legs), sometimes with a spurious forward chain.

## Approach

Fail-safe, **non-destructive** first layer (v1) — it does not touch stock automatically:

- After `button_cancel_remaining`, detect the phantom pattern (inverted move Customer->internal, alive, `to_refund`, no `origin_returned_move_id`) and raise a **warning activity** on the order.
- The order form shows a **warning banner** (an `alert`, not a button in the action box) with a **"Sanitize phantom counter-deliveries (beta)"** button. Restricted to **Inventory Administrators** (`stock.group_stock_manager`), enforced in the view and in the method (`AccessError`).
- **Sanitize cancels the full phantom footprint** in the procurement group, not just the entry move: the whole `to_refund` reverse chain (Customer->Despacho + internal put-back legs) plus, for phantom products with no remaining demand, the spurious forward chain. **Genuine returns** (`origin_returned_move_id`) and **products with pending demand** are never touched.
- A **search filter** ("With Phantom Counter-deliveries") lists affected orders. Not a stored field: a `search` method runs a single indexed query on `stock.move`.
- **Beta opt-in:** activity, banner button and filter are gated behind the system parameter `sale_stock_ux.sanity_pickings` (off by default).

Prevention was deliberately avoided: reimplementing the routing proved fragile across multi-step scenarios (reverted before; the OCA `sale_order_line_cancel_sale_stock` prevention module is itself beta and warns against confirmed-line qty edits). This reacts to the artifact core already produced, so it does not re-validate per Odoo version.

## Known limitations (beta)

- A phantom that kept an `origin_returned_move_id` (born linked to an original outgoing move) is intentionally not flagged, to avoid touching genuine returns.
- Prevention of new phantoms is out of scope (the known route triggers are already prevented in `stock_ux@origin/18.0`; this covers residual/legacy/unknown cases).
- Sale side only; purchase mirror to follow.

## Test plan

`tests/test_phantom_counterdelivery.py` — reactive pipeline on the phantom artifact:
- no false positive on a clean order; detects + single (idempotent) warning activity
- beta gate off -> banner hidden, no activity, filter matches nothing
- search filter finds/excludes the order (per beta flag)
- sanitize requires Inventory Administrator (AccessError otherwise)
- genuine return (`origin_returned`) not flagged
- **full multi-leg footprint**: reverse chain (DEVC + internal put-back legs) all cancelled; genuine return kept; forward spurious cancelled only when the product has no remaining demand; forward kept when demand pending

`tests/test_phantom_counterdelivery_e2e.py` — real cancel-remaining across 1/2/3-step routes (standard pull, pull+push, pull+push+push, push_domain): detector agrees with the real chain and stays quiet on healthy cancels and legitimate internal put-backs.

All green locally (16 tests).

Task: https://www.adhoc.inc/odoo/project.task/73048

### DIFF
--- a/sale_stock_ux/models/sale_order.py
+++ b/sale_stock_ux/models/sale_order.py
@@ -3,7 +3,8 @@
 # directory
 ##############################################################################
 from odoo import _, api, fields, models
-from odoo.exceptions import UserError
+from odoo.exceptions import AccessError, UserError
+from odoo.tools.float_utils import float_compare
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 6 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1846)

PR #1848, opened 2026-09-08
[ADD] sale_stock_ux: test cancel-remaining does not refund internal put-backs

jcadhoc:
### What

Regression test for the companion fix in **stock_ux** (`ingadhoc/stock#1032`, same branch name → same runbot build).

### Scenario

A multi-step push delivery (pull + push + push) with a **partial delivery** whose remainder is advanced one step, then the remaining quantity is cancelled. Before the fix the internal put-back (`Packing Zone -> Stock`) was left tagged `to_refund`, inflating returned quantities and invoicing status.

The test asserts no `internal -> internal` move ends up flagged as a refund after cancelling the remaining. It also covers the plain partial-delivery and the partial-delivery-plus-return variants (the genuine customer return keeps `to_refund`, as expected).

### DIFF
--- a/sale_stock_ux/tests/__init__.py
+++ b/sale_stock_ux/tests/__init__.py
@@ -3,3 +3,4 @@ from . import test_kit_return_uom
 from . import test_return_of_return
 from . import test_stock_info_batching
 from . import test_invoice_status_return
+from . import test_cancel_remaining_partial_return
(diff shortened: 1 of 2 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1848)

PR #1849, opened 2026-09-08
[REF] sale_order_type_ux: drop downpayment account override, now in account_multicompany_ux

lef-adhoc:
## What

`sale_order_type_ux` overrode `_prepare_invoice_line` to force the downpayment account of the order company on the invoice line. It is needed because `_create_invoices` creates the invoice in the order company and only afterwards moves it with the change company wizard, while `sale._prepare_invoice_line` copies the account of the already invoiced downpayment without checking its company.

That problem is not specific to sale order types: any database with more than one company hits it as soon as a downpayment invoice ends up in another company than its order, and databases without this module have no fix at all today.

## How

The guard now lives in `account_multicompany_ux` (ingadhoc/multi-company#320), which this module already depends on, so this copy is redundant and is removed.

Same target company (the order one) and same criterion (downpayment account of the product category, income account otherwise), with two intentional differences:

- the account is looked up on the line product first and on the order products only as a fallback;
- an account shared by a parent company is accepted instead of being replaced, which is what `_check_company` actually allows (`check_companies_domain_parent_of`).

The `check_company=False` relaxation on `tax_id` stays — it covers the taxes, not the account.

## Test plan

Depends on ingadhoc/multi-company#320: both branches share the name, so runbot builds them together and this module keeps its behaviour only with that one in.

1. Sale order type whose journal belongs to another company.
2. Sale order in company B with that type, confirmed.
3. Invoice → Down payment. The invoice is created in B and moved to A by the wizard. Post it.
4. Back on the order → Create invoice with "Deduct down payments".

The final invoice is created and the deduction line takes an account of the order company, same as before this change.

Internal ticket: https://www.adhoc.inc/odoo/helpdesk.ticket/127169

### DIFF
--- a/sale_order_type_ux/models/sale_order_line.py
+++ b/sale_order_type_ux/models/sale_order_line.py
@@ -11,27 +11,3 @@ class SaleOrderLine(models.Model):
     # Quitamos el check_company ya que no permitira
     # realizar un cambio de compañia entre la venta y su factura de anticipo
     tax_id = fields.Many2many(check_company=False)
-
-    def _prepare_invoice_line(self, **optional_values):
-        res = super()._prepare_invoice_line(**optional_values)
-        # Fix multicompañía:
-        # Cuando se cambia la compañía de una factura de anticipo (ya sea con el wizard de "change company"
-        # o mediante el sale order type), Odoo reutiliza la cuenta contable de esa factura de anticipo
-        # para las siguientes facturas.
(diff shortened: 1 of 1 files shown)


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1849)

maq-adhoc:
@roboadhoc override=ci/runbot-oba

maq-adhoc:
@roboadhoc r+ nobump

PR #1860, opened 2026-09-15
[FIX] portal_sale_distributor: read access to the stock models mrp_subcontracting restricts

gal-adhoc:
## What

Four record rules for the portal distributor group, on `stock.location`, `stock.picking.type`, `stock.move` and `stock.move.line`.

## Why

A portal distributor is also in `base.group_portal`. `mrp_subcontracting` adds rules on that group tying those models to what belongs to the subcontracting of the partner — its pickings, its BoMs, its productions. A distributor has none of that, so the rules match nothing and reading a record raises:

```
Estos registros están restringidos.
<user> no tiene acceso 'leer' a:
- Localización de Inventarios (stock.location)
```

It shows up as soon as the distributor tries to add a line to a sale order.

The `ir.model.access` entries for these models are already granted to the group, so an ACL is not what is missing: what blocks is the record rule. The module already knew this pattern — see the comment above `portal_stock_warehouse_rule` — and countered it for `stock.warehouse` only.

`stock.picking` is not in the list on purpose: the portal rule on transfers already covers the orders of the partner, and group rules are OR'ed.

## On the domain

Same shape 19.0 gave `portal_stock_warehouse_rule` in 348f41f5, after the review on #1637 pointed out that a rule with `[(1, '=', 1)]` is not a rule:

```
['|', ('company_id', '=', False), ('company_id', 'in', user.company_ids.ids)]
```

It neutralises the subcontracting rule, which is the point, while still carrying a criterion of its own and staying on top of the global multi-company rule of stock.

## Test plan

New test in `portal_sale_distributor/tests/`, skipped when `mrp_subcontracting` is not installed and run on the builds where it is.

Verified on a local database with `portal_sale_distributor` + `mrp_subcontracting`:

- with the rules, a distributor reads the four models;
- deactivating the four rules, the four fail again with the access error above;
- `0 failed, 0 error(s)`.

## Port to 19.0

Not automatic. 19.0 dropped `mrp_subcontracting`'s uom rule along with the units of measure rework, so before porting, each of these four models has to be checked against 19.0 the same way: if nothing restricts it any more, the rule is dead weight.

Ref: https://www.adhoc.inc/odoo/helpdesk.ticket/127896


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1860)

lef-adhoc:
@roboadhoc r+ bump

PR #1863, opened 2026-09-17
[FIX] sale_delivery_ux: set delivery as done only on outgoing picking

lef-adhoc:
## What happens

With a 2 or 3 steps delivery route (PICK → PACK → OUT), validating the first internal transfer already sets `qty_delivered = 1` on the delivery line of the sale order. Core `sale_stock` then computes the order as **Partially Delivered**, even though nothing left the warehouse: the delivery line shows 1.00 delivered while every product line is still at 0.00. With a single step route the difference is not visible.

## Cause

Since the hook moved from `_add_delivery_cost_to_so` to `_action_done` (#1487), the delivery line is completed on **any** validated picking, with no check on whether that picking is the one going out to the customer. That move was needed because `_add_delivery_cost_to_so` only runs from `send_to_shipper`, so a fixed price carrier never reached it — this PR keeps that behaviour.

## Fix

Restrict the completion to pickings having at least one outgoing move, reusing the criteria core already applies in `sale_stock` (`stock.move._is_outgoing()`: destination is a customer/supplier location, or a transit location without company). `location_dest_id` is used on purpose — `location_final_id` would match the internal PICK again under the push mechanism.

As a side effect it also fixes a second issue in the same method: `_action_done` can run over a recordset of pickings, and `self.sale_id.order_line` reached the orders of every picking involved, so an unrelated internal transfer completed the delivery line of another order.

## Tests

The module had no tests. Added `tests/test_delivery_line_qty_delivered.py`:

- `test_three_steps_delivery_marked_on_last_step`: PICK and PACK leave the delivery line at 0 and the order as `started`; validating OUT sets it to 1 and the order to `full`.
- `test_one_step_delivery_marked_on_validation`: non regression, a single step route still completes the delivery line.
- `test_internal_picking_does_not_mark_another_order`: `_action_done` over an internal picking of one order plus an outgoing picking of another one only completes the second one.

Checked on a local 19.0 database: the first and third tests fail without the fix and pass with it; the single step one passes either way. The 18.0 run is left to runbot.

Suggested policy: `nobump` — Python logic only, no field definitions, views or data changed.

Task: https://www.adhoc.inc/odoo/project.task/75193


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1863)

gal-adhoc:
@roboadhoc r+ nobump

PR #1864, opened 2026-09-18
[ADD] sale_exception_compat: keep the previous exception flow

lef-adhoc:
## What

New module `sale_exception_compat`. It restores the exception flow `sale_exception` had before OCA/server-tools#3590 and OCA/sale-workflow#4255, so those repositories can be updated without changing the behaviour our modules rely on.

## Why

Upstream now stores the detected exceptions through a **second database cursor** that commits, and raises `BaseExceptionError` so the ongoing transaction is rolled back. That second connection writes the same rows the ongoing transaction is about to write, which breaks in two ways:

- **Serialization failure.** The wizard flow (tick "Ignore Exceptions", confirm) writes `ignore_exception` and then confirms. The second cursor commits on that row, and the main transaction — started before that commit, under REPEATABLE READ — fails on its final flush with `could not serialize access due to concurrent update`. Reproduced 3 out of 3; the order is never confirmed.
- **Blocked worker.** When the main transaction takes the row lock first, the second cursor waits on it while the main one waits for the second cursor to finish. Postgres does not see the cycle, so it does not abort: the process hangs with no timeout (killed after 10 minutes in testing).

On top of that, the error handler that turns `BaseExceptionError` into the popup is only registered in `web.assets_backend`, so portal, e-commerce, API and cron callers get a raw error where they used to get a normal response.

## How

- `base.exception.method` — replaces `detect_exceptions()` with the previous one: writes in the ongoing transaction, returns the rules, no raise.
- `sale.order` — restores `action_confirm()` with the popup, and keeps the `_register_hook` that puts the detection outermost in the MRO.
- `sale.order.line` — skips the `sale_exception` override, which opens the second cursor as well.

The rollback now lives in a single method used by both `action_confirm()` and the registry hook, and clears the environment cache afterwards. Previously the hook returned the popup before reaching the rollback, and the dirty cache wrote back what the rollback had undone — so writes other modules made before the detection survived. They no longer do.

`auto_install` is on, so it follows `sale_exception`.

## Test plan

Against a database with `sale_exception_print`, `portal_sale_distributor` and `website_sale_force_confirmation` installed, running the updated OCA code:

- Confirm an order matching a rule → exception popup, order stays in draft, exceptions stored.
- A write made before the detection → rolled back (it was not, before this module).
- Wizard "Ignore Exceptions" + confirm → order is confirmed.
- Write on the order, flush, then confirm → popup, no hang.
- `action_quotation_send` with a `block_print` rule → print popup.
- `action_confirm_distributor` with an exception → one chatter message, order in draft.
- Portal accept and sign → `sign_ok`, no error shown to the customer.
- `call_kw` on `action_confirm` → returns the wizard action, not an error.
- `test_all_draft_orders()` → returns True.
- `-u base_exception,sale_exception,sale_exception_compat` → no hang.

## Follow-up

The two upstream problems should be reported to OCA. This module is meant to be dropped once upstream stops writing through the second cursor.


roboadhoc:
[Pull request status dashboard](https://mergebot.adhoc.ar/ingadhoc/sale/pull/1864)

maq-adhoc:
@roboadhoc r+ bump

